PR #18, opened 2023-04-04
Export relationships query fix

simsta6:
- Modified query for getting all relationships. Now it selects all columns and does not return invalid relationships that do not point to non existing elements.
- Added better support for VS Code IDE

### DIFF
--- a/packages/transformer/.vscode/launch.json
+++ b/packages/transformer/.vscode/launch.json
@@ -10,11 +10,11 @@
                 "800000",
                 "--require",
                 "source-map-support/register",
-                "${workspaceFolder}\\lib\\cjs\\test\\**\\*.test.js"
+                "${workspaceFolder}/lib/cjs/test/**/*.test.js"
             ],
             "internalConsoleOptions": "openOnSessionStart",
             "name": "Mocha Tests",
-            "program": "C:\\Users\\Simas.Stasys\\repos\\transformer\\packages\\transformer\\node_modules\\mocha\\bin\\_mocha",
+            "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
             "request": "launch",
             "skipFiles": [
                 "<node_internals>/**"
--- a/packages/transformer/.vscode/launch.json
+++ b/packages/transformer/.vscode/launch.json
@@ -5,21 +5,13 @@
     "version": "0.2.0",
     "configurations": [
         {
-            "args": [
-                "--timeout",
-                "800000",
-                "--require",
-                "source-map-support/register",
-                "${workspaceFolder}/lib/cjs/test/**/*.test.js"
-            ],
-            "internalConsoleOptions": "openOnSessionStart",
             "name": "Mocha Tests",
-            "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
+            "type": "node",
             "request": "launch",
-            "skipFiles": [
-                "<node_internals>/**"
-            ],
-            "type": "node"
+            "runtimeExecutable": "npm",
+            "runtimeArgs": [
+                "run-script", "test"
+            ]
         }
     ]
 }
--- a/packages/transformer/.vscode/launch.json
+++ b/packages/transformer/.vscode/launch.json
@@ -1,0 +1,25 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "args": [
+                "--timeout",
+                "800000",
+                "--require",
+                "source-map-support/register",
+                "${workspaceFolder}\\lib\\cjs\\test\\**\\*.test.js"
+            ],
+            "internalConsoleOptions": "openOnSessionStart",
+            "name": "Mocha Tests",
+            "program": "C:\\Users\\Simas.Stasys\\repos\\transformer\\packages\\transformer\\node_modules\\mocha\\bin\\_mocha",
+            "request": "launch",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "type": "node"
+        }
+    ]
+}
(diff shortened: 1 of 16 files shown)


packages/transformer/.vscode/launch.json
```diff
@@ -0,0 +1,25 @@
...
+            "args": [
+                "--timeout",
+                "800000",
+                "--require",
+                "source-map-support/register",
+                "${workspaceFolder}/lib/cjs/test/**/*.test.js"
+            ],
+            "internalConsoleOptions": "openOnSessionStart",
+            "name": "Mocha Tests",
+            "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
```
MichaelBelousov, line 17: maybe it's worth just having this run `npm run test` so we don't need to keep the two debugging script in sync if we change mocha configuration?

packages/transformer/src/IModelExporter.ts
```diff
@@ -685,12 +685,12 @@ export class IModelExporter {
...
     Logger.logTrace(loggerCategory, `exportRelationships(${baseRelClassFullName})`);
-    const sql = `SELECT ECInstanceId FROM ${baseRelClassFullName}`;
+    const sql = `SELECT ECInstanceId, ECClassId, * FROM ${baseRelClassFullName}`;
     await this.sourceDb.withPreparedStatement(sql, async (statement: ECSqlStatement): Promise<void> => {
       while (DbResult.BE_SQLITE_ROW === statement.step()) {
-        const relInstanceId: Id64String = statement.getValue(0).getId();
-        const relProps: RelationshipProps = this.sourceDb.relationships.getInstanceProps(baseRelClassFullName, relInstanceId);
-        await this.exportRelationship(relProps.classFullName, relInstanceId); // must call exportRelationship using the actual classFullName, not baseRelClassFullName
+        const relationshipId = statement.getValue(0).getId();
+        const relationshipClass = statement.getValue(1).getClassNameForClassId();
```
MichaelBelousov, line 692: thank you for fixing that, I was planning on removing that completely overkill `relationships.getInstanceProps` during other work. Resolve this at will.

packages/transformer/src/test/standalone/IModelExporter.test.ts
```diff
@@ -73,4 +73,36 @@ describe("IModelExporter", () => {
...
+        classFullName: GraphicalElement3dRepresentsElement.classFullName,
+        targetId: "",
+        sourceId: "",
+      };
+
+      sourceDb.relationships.insertInstance(relationshipProps);
+      sourceDb.saveChanges();
+
+      const sourceRelationships = new Array<any>();
+      sourceDb.withStatement("SELECT ECInstanceId FROM bis.ElementRefersToElements", (stmt) => sourceRelationships.push(...stmt));
```
MichaelBelousov, line 92: nit: you can just do:
```ts
const sourceRels = sourceDb.withStatement(sql, (s) => [...s]);
```
since `ECSqlStatement` has an iterator definition

packages/transformer/src/test/standalone/IModelExporter.test.ts
```diff
@@ -73,4 +73,36 @@ describe("IModelExporter", () => {
 
     sourceDb.close();
   });
+
+  describe("exportRelationships", () => {
+    it("should not export relationships that do not have source or target elements", async () => {
+      const sourceDbPath = IModelTransformerTestUtils.prepareOutputFile("IModelExporter", "InvalidRelationship.bim");
+      const sourceDb = SnapshotDb.createEmpty(sourceDbPath, { rootSubject: { name: "invalid-relationships" } });
+
+      const relationshipProps: RelationshipProps = {
```
MichaelBelousov, line 82: can you add to this test:
1. a relationship with valid ids, where you insert two elements, create a relationship between them, and then delete one of the elements?
2. a relationship with valid source, invalid target
3. a relationship with invalid source, valid target

(it would be nice btw if you posted that message you sent to me as the PR description). If I understand the source problem you sent to me correctly, you said the "other element is non existent" in the invalid relationships you're seeing, which means I think we need to test that directly, aka case 1 above.

Ideally I would actually like us to transform the invalid relationships in case 1 but nullify the id they point to, based on whether the transformer options' `danglingPredecessorsBehavior` is set to reject or not... thoughts?

MichaelBelousov: to add to this, if you personally want your transformer implementation to ignore such "invalid" relationships, you can do:

```ts
public override shouldExportRelationship(sourceRel) {
  return Id64.isValid(sourceRel.sourceId) && Id64.isValid(sourceRel.targetId);
}
```

MichaelBelousov: Since it doesn't work anyway at the moment, I have put up #21 to fix this in the future. I will keep the comment open because the list of test cases I would like you to add still stand.

packages/transformer/src/test/standalone/IModelExporter.test.ts
```diff
@@ -73,4 +73,69 @@ describe("IModelExporter", () => {
...
+        // only source element is invalid
+        {
+          classFullName: GraphicalElement3dRepresentsElement.classFullName,
+          targetId: physicalObject4,
+          sourceId: "",
+        },
+      ];
+
+      invalidRelationshipsProps.forEach((props) => sourceDb.relationships.insertInstance(props));
+      sourceDb.withSqliteStatement(`DELETE FROM bis_Element WHERE Id = ${parseInt(physicalObject1, 16)}`, (stmt) => stmt.next());
```
MichaelBelousov, line 123: can you use `deleteElementTree` from `@itwin/core-backend`? This will leave entries in the `bis_GeometricElement3d` table and is not the right way to delete this.

MichaelBelousov (changes requested):

packages/transformer/src/test/standalone/IModelExporter.test.ts
simsta6: In this way, it deletes a relationship too (as it should be). So the issue, that connectors introduce, is not replicated in this way

MichaelBelousov: my bad, try `sourceDb.elements.deleteElement`

MichaelBelousov (approved):

PR #30, opened 2023-05-01
add custom prerelease prefix option to workflow

MichaelBelousov:
add custom prerelease prefix option to workflow


MichaelBelousov:
done in #32

PR #35, opened 2023-05-15
Added additional args to Exporter.exportChanges

simsta6:
- Added additional args to Exporter.exportChanges
- Added deprecation tag for old exportChanges overload
- Added deprecation tag for old IModelTransformer.transformChanges overload


packages/transformer/src/IModelExporter.ts
```diff
@@ -81,7 +81,7 @@ export abstract class IModelExportHandler {
    *       This will become a part of onExportElement once that becomes async
    * @internal
    */
-  public async preExportElement(_element: Element): Promise<void> {}
+  public async preExportElement(_element: Element): Promise<void> { }
```
MichaelBelousov, line 84: unnecessary change? Where does this come from? Was this a linter rule?

packages/transformer/src/IModelExporter.ts
```diff
@@ -179,11 +179,18 @@ export class IModelExporter {
   private _sourceDbChanges?: ChangedInstanceIds;
   /**
    * Retrieve the cached entity change information.
-   * @note This will only be initialized after [IModelExporter.exportChanges] is invoked.
+   * @note This will be initialized after [IModelExporter.exportChanges] is invoked or it can be set.
```
MichaelBelousov, line 182:
```suggestion
   * @note This may be set, or otherwise will be initialized after [IModelExporter.exportChanges] is invoked.
```

packages/transformer/src/IModelExporter.ts
```diff
@@ -751,7 +760,7 @@ export class IModelExporter {
    * You may override this to load arbitrary json state in a transformer state dump, useful for some resumptions
    * @see [[IModelTransformer.loadStateFromFile]]
    */
-  protected loadAdditionalStateJson(_additionalState: any): void {}
+  protected loadAdditionalStateJson(_additionalState: any): void { }
```
MichaelBelousov, line 763: is this the linter?

MichaelBelousov:
what is the purpose of this change? What is a use case for setting it?

packages/transformer/src/IModelExporter.ts
simsta6: I formatted this file from vs code, so it may be the linter

simsta6:
> what is the purpose of this change? What is a use case for setting it?

In our service, we need `ChangedInstanceIds` data. To reuse this data, I decided to add a public getter because adding a hook for this is not worth it.

MichaelBelousov:
> In our service, we need `ChangedInstanceIds` data. To reuse this data, I decided to add a public getter because adding a hook for this is not worth it.

Maybe you can make it an option like `sourceChangeData`? I'd rather not expose a setter that can be used at any time.

packages/transformer/src/IModelExporter.ts
```diff
@@ -268,9 +268,11 @@ export class IModelExporter {
    * @param user The user
    * @param startChangesetId Include changes from this changeset up through and including the current changeset.
    * If this parameter is not provided, then just the current changeset will be exported.
+   * @param changedInstanceIds Instance class that contains modified elements between 2 versions of an iModel.
+   * If this parameter is not provided, then [[ChangedInstanceIds.initialize]] will be called to discover changed elements.
    * @note To form a range of versions to export, set `startChangesetId` for the start (inclusive) of the desired range and open the source iModel as of the end (inclusive) of the desired range.
    */
-  public async exportChanges(user?: AccessToken, startChangesetId?: string): Promise<void> {
+  public async exportChanges(user?: AccessToken, startChangesetId?: string, changedInstanceIds?: ChangedInstanceIds): Promise<void> {
```
MichaelBelousov, line 275: instead of adding a third plain argument, can it be an object containing this, so that future additional arguments can be added in a non-breaking way? Eventually I intend to deprecate the overload passing a string argument for `startChangesetId` and instead also put that in the object

MichaelBelousov: It's not essential but yes, since it still introduces the new argument

simsta6: We should start doing this as a standard practice

MichaelBelousov: Yeah I'm willing to always bump minor on API changes, including non-breaking ones. It is better form.

simsta6: Also converted processChanges to use object as argument

MichaelBelousov: what you have now is a breaking change. I meant the following:

```ts
 public async exportChanges(user?: AccessToken, startChangesetId?: string, options: ExportChangesArgs): Promise<void> {
```

And then later we can deprecate that overload and do this. Although you can do so in this PR too, and we might as well if we're making it a minor release.

```ts
 /** @deprecated Don't use this overload */
 public async exportChanges(user?: AccessToken, startChangesetId?: string, options: ExportChangesArgs): Promise<void>
 /** desc */
 public async exportChanges(options: ExportChangesArgs): Promise<void>;
 public async exportChanges(userOrOptions?: AccessToken | ExportChangesArgs, startChangesetId?: string, options?: ExportChangesArgs): Promise<void> {
   const options = typeof userOrOptions === "string"
     ? {
       user: userOrOptions,
       startChangesetId,
       ...options
     }
     : userOrOptions; 
   //...
```

simsta6: Great idea!

MichaelBelousov: could you actually go ahead and add the deprecated overload as I showed above?

simsta6: Added

MichaelBelousov: do whatever works for VSCode, e.g. hovering later usages of `exportChanges` should show the description, but the invalid overload should be detected by both typescript (showing a ~strikethrough~ in vscode ) and eslint.

I believe it's really what I posted above. The full description remains on the valid overload signature which is declared first, then the deprecated overload is declared second with no description except `@deprecated in 0.1.x, use a single ExportChangesArgs object instead`, and that should work. Unless I removed them, I believe I've done this in other places.

simsta6: I changes descriptions. I was using itwinjs core lib as example.

MichaelBelousov: do you think you could cherry-pick in my version where I did something similar? 2f1189eafa68c7382c7e220030961157f487599a
You might also need c2f2da9d400f34e7b8b1c6b2fcd5d07fdae537a4...

simsta6: Is these changes still not in master? Maybe it would be better just merge this PR or close it and start over again

simsta6: There is multiple issues with those commits:
 - using types like `ChangesetIndexOrId` makes hardly readable code
 - importing types from different file, that can be reused, but actually is for different method is also makes code look like spaghetti

I would rather stick with my current version

MichaelBelousov: well the types are highly related, those functions call each other so they should have a relatively consistent API. And I disagree that `ChangesetIndexOrId` makes for hardly readable code. The value of using `ChangesetIndexOrId` is that the user can pass what they have without having to resolve the other kind using a query. This change actually reduced the amount of querying the transformer does which needs an index but was requesting an id.

If you're referring to the new "args" declarations as spaghetti, they are slightly unfortunate but that's a natural reality of supporting multiple overloads and makes sense until the next major version when we can drop the deprecated overload.

Are you still against this?

simsta6: I would rather have multiple types or interfaces which extends basic ones. Because there is a possibility to have a need to add additional property that will be used in one function and not in other function

MichaelBelousov: maybe I forgot a commit, there should be a second type that extends from the base.

MichaelBelousov: I think I forgot [ff19036f](https://github.com/iTwin/imodel-transformer/commit/ff19036ec6456c59891f6d56bdb2982e082d8d2f)
It's really a subset of federation-guid-optimization branch, which you could extract using
`git add -p federation-guid-optimization packages/transformer/src/IModelTransformer.ts packages/transformer/src/IModelExport.ts` and just ignore all other changes. Or try to cherry-pick those 3 commits but I didn't thoroughly check if that contains all the changes.

MichaelBelousov: I created #55 to demonstrate what I'd prefer

packages/transformer/src/IModelExporter.ts
```diff
@@ -264,29 +278,48 @@ export class IModelExporter {
...
-   */
-  public async exportChanges(user?: AccessToken, startChangesetId?: string): Promise<void> {
-    if (!this.sourceDb.isBriefcaseDb()) {
+  /**
+   * Export changes from the source iModel.
+   */
+  public async exportChanges(args?: ExportChangesOptions): Promise<void>;
+  /** @deprecated in 0.1.x, use a single ExportChangesArgs object instead */
+  public async exportChanges(accessToken?: AccessToken, startChangesetId?: string, args?: ExportChangesOptions): Promise<void>;
+  /** @internal Don't use this overload. */
```
MichaelBelousov, line 287: nit: is it necessary to mark this as internal? It's not an overload but the implementation. I think typescript will identify similar usage (more than 1 argument) as the deprecated overload so this is unnecessary

MichaelBelousov (approved):
need to fix lint errors but LGTM!

packages/transformer/src/IModelExporter.ts
simsta6: Okay, I removed it :)

PR #51, opened 2023-05-30
extend branch workflow tests to include relationship delete

MichaelBelousov:
extend branch workflow tests to include relationship delete

This appears to be broken in main so leaving a draft and merging into the fed guid work where it needs to be redone anyway.


MichaelBelousov:
now a part of ongoing fedguid work

PR #53, opened 2023-05-31
Regression test pipeline

DanRod1999:
Originally created this branch for the yaml pipeline, however now it holds a refactored test file for the identity transform test.


packages/performance-tests/README.md
```diff
@@ -0,0 +1,51 @@
+# Presentation Performance Tests
+
+A package containing performance tests for the [`@itwin/transformer` library](../../README.md).
```
MichaelBelousov, line 3:
```suggestion
A package containing performance tests for the [`@itwin/imodel-transformer` library](../../README.md).
```

packages/performance-tests/README.md
```diff
@@ -0,0 +1,51 @@
...
+  transform the iModel, editing geometry as we go using elementGeometryBuilderParams
+- *Optimistically Locking Remote Target*
+- *Pessimistically Locking Remote Target*
+- *Processing Changes*
+- *More Branching Stuff*
+
+<!-- FIXME: this isn't true yet-->
+All tests run in isolation on every iModel in these projects:
+
+- https://qa-connect-imodelhubwebsite.bentley.com/Context/892aa2c9-5be8-4865-9f37-7d4c7e75ebbf
```
MichaelBelousov, line 23: remove these references to internal imodels. They should be configured in an .env file or something, that we can set in our internal pipelines, and external users should also be able to configure their own imodels for these tests.

packages/performance-tests/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "private": true,
+  "license": "MIT",
+  "version": "0.1.0",
+  "scripts": {
+    "build": "tsc 1>&2",
+    "build:ci": "npm run -s build",
+    "clean": "rimraf lib",
+    "lint": "eslint --max-warnings 0 \"./src/**/*.ts\" 1>&2",
+    "test": "mocha --delay --timeout 300000 --require ts-node/register test/**/*.test.ts",
+    "test-reg": "mocha --delay --timeout 300000 --require ts-node/register test/**/TransformerRegression.test.ts",
```
MichaelBelousov, line 12: can you update this and its references to `test-regression`? `reg` is kind of hard to understand without reading the rest of the line

packages/performance-tests/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "test-mocha": "mocha --delay --timeout 300000 \"./lib/**/TransformerRegression.test.js\"",
+    "process-reports": "node scripts/process-reports"
+  },
+  "repository": {},
+  "dependencies": {
+    "@itwin/core-backend": "3.6.0 - 3.6.1",
```
MichaelBelousov, line 18: these dependencies are not aligned with those in `main`. Please update them to the ones in main, which should be something like `3.6.0 - 3.7.999`, don't forget dev deps

packages/performance-tests/test/TestContext.ts
```diff
@@ -0,0 +1,84 @@
...
+export interface TestIModel {
+  name: string;
+  iModelId: string;
+  iTwinId: string;
+  tShirtSize: string;
+  load: () => Promise<BriefcaseDb>;
+}
+
+export const testITwinIds = [
+  "892aa2c9-5be8-4865-9f37-7d4c7e75ebbf",
```
MichaelBelousov, line 23: let's move this to some sort of configuration file and not include literal ids here

packages/performance-tests/test/TestContext.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+}
+
+export async function downloadAndOpenBriefcase(briefcaseArg: Omit<RequestNewBriefcaseArg, "accessToken">): Promise<BriefcaseDb> {
+  const PROGRESS_FREQ_MS = 2000;
+  let nextProgressUpdate = Date.now() + PROGRESS_FREQ_MS;
+
+  assert(IModelHost.authorizationClient !== undefined, "auth client undefined");
+  const briefcaseProps =
+    BriefcaseManager.getCachedBriefcases(briefcaseArg.iModelId)[0] ??
```
MichaelBelousov, line 65: I just ran into an issue with this code in the place from which I copied it, not that it matters that much in our current tests, but the cached iModel may not match the `briefcaseArg`, specifically which changeset it's on. You should probably try to find a cached briefcase with the same changeset id, and if not, always download.

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+import { TestBrowserAuthorizationClient } from "@itwin/oidc-signin-tool";
+import { Reporter } from "@itwin/perf-tools";
+import * as os from "os";
+
+async function sleep(ms: number) {
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms);
+  });
+}
+const testCasesMap = new Map<string, any>();
```
MichaelBelousov, line 34: nit: why not an inline map?

```ts
const testCasesMap = new Map([
  ["identity transform", require("./identity-transformer")],
]);
```

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+testCasesMap.set("identity transform", require("./identity-transformer"));
+// testCasesMap.set("prepare fork transform", require("./prepare-fork"));
+
+const loggerCategory = "Transformer Performance Tests Identity";
+const outputDir = path.join(__dirname, ".output");
+
+const setupTestData = async () => {
+  const logLevel = +process.env.LOG_LEVEL!;
```
MichaelBelousov, line 42: can you do like `+(process.env.LOG_LEVEL ?? "2")`?

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    ? new TestBrowserAuthorizationClient({
+      clientId: process.env.OIDC_CLIENT_ID,
+      redirectUri: process.env.OIDC_REDIRECT,
+      scope: "itwins:read imodels:read imodels:modify",
+      authority: "https://qa-ims.bentley.com",
+    }, user)
+    : new NodeCliAuthorizationClient({
+      clientId: process.env.OIDC_CLIENT_ID,
+      redirectUri: process.env.OIDC_REDIRECT,
+      scope: "imodelaccess:read storage:modify realitydata:read imodels:read library:read imodels:modify realitydata:modify savedviews:read storage:read library:modify itwinjs savedviews:modify",
```
MichaelBelousov, line 83: do we not want these in the .env file?

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        describe(`Transforms of ${iModel.name}`, async () => {
+          testCasesMap.forEach(async (testCase, key, _map) => {
+            it(key, async () => {
+              reporter = await testCase.default(iModel, os, reporter);  // add timeout(0)
+            }).timeout(0);
+          });
+        });
+      }
+    });
+  });
```
MichaelBelousov, line 120: space between callbacks?

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+};
+
+void (async function () {
+  const testIModels = await setupTestData();
+  var reporter = new Reporter();
+  const reportPath = initOutputFile("report.csv", outputDir);
+
+  // could probably add an outer describe here (surounding both foreaches) for any before after logic that isn't related to getting the test data. but idk for sure 
+  describe('Transformer Regression Tests', function () {
+    testIModels.forEach(async (value, index, _array) => {
```
MichaelBelousov, line 108: why not just do `for (const iModel of getTestIModels())` here? Does mocha dislike it?
Also can you remove the unused `_array` reference? You don't need to handle optional callback parameters

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  await authClient.signIn();
+
+  const hostConfig  = new IModelHostConfiguration();
+  hostConfig.authorizationClient = authClient;
+  const hubClient = new IModelsClient({ api: { baseUrl: `https://${process.env.IMJS_URL_PREFIX}api.bentley.com/imodels` } });
+  hostConfig.hubAccess = new BackendIModelsAccess(hubClient);
+  await IModelHost.startup(hostConfig);
+
+  const arrImodels = [];
+  for await (const iModel of getTestIModels()) {
```
MichaelBelousov, line 95: this sort of defeats the purpose of that iterator? Might as well rewrite getTestIModels?

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  var reporter = new Reporter();
+  const reportPath = initOutputFile("report.csv", outputDir);
+
+  // could probably add an outer describe here (surounding both foreaches) for any before after logic that isn't related to getting the test data. but idk for sure 
+  describe('Transformer Regression Tests', function () {
+    testIModels.forEach(async (value, index, _array) => {
+      const iModel = value;
+      if (index === 2 || index === 5) {
+        describe(`Transforms of ${iModel.name}`, async () => {
+          testCasesMap.forEach(async (testCase, key, _map) => {
```
MichaelBelousov, line 112: remove unused `_map` param

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+void (async function () {
+  const testIModels = await setupTestData();
+  var reporter = new Reporter();
+  const reportPath = initOutputFile("report.csv", outputDir);
+
+  // could probably add an outer describe here (surounding both foreaches) for any before after logic that isn't related to getting the test data. but idk for sure 
+  describe('Transformer Regression Tests', function () {
+    testIModels.forEach(async (value, index, _array) => {
+      const iModel = value;
+      if (index === 2 || index === 5) {
```
MichaelBelousov, line 110: can you use a more self-documenting identifier for the iModels instead of their index? Like matching on names? Also maybe we can move those identifiers into the .env of some other external test configuration since it won't make sense for external users trying to run these tests?

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+            }).timeout(0);
+          });
+        });
+      }
+    });
+  });
+  after(async () => {
+    reporter.exportCSV(reportPath);
+  });
+  // See 'DELAYED ROOT SUITE' on https://mochajs.org/
```
MichaelBelousov, line 124: use anchored link
```suggestion
  // See 'DELAYED ROOT SUITE' on https://mochajs.org/#delayed-root-suite
```

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  // This function is a special bcallback function provided by mocha when passing it the --delay flag. This gives us an opportunity to load in the iModels that we'll be testing so we can dynamically generate testcases.
```
MichaelBelousov, line 125:
```suggestion
  // This function is a special callback function provided by mocha when passing it the --delay flag. This gives us an opportunity to load in the iModels that we'll be testing so we can dynamically generate testcases.
```

packages/performance-tests/test/identity-transformer.ts
```diff
@@ -0,0 +1,79 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import * as path from "path";
+import * as fs from "fs";
+import { Element, IModelHost, IModelHostConfiguration, Relationship, SnapshotDb, BriefcaseDb } from "@itwin/core-backend";
+import { Logger, LogLevel, PromiseReturnType, StopWatch } from "@itwin/core-bentley";
+import { IModelTransformer, TransformerLoggerCategory } from "@itwin/imodel-transformer";
+//import { TestBrowserAuthorizationClient } from "@itwin/oidc-signin-tool";
```
MichaelBelousov, line 10: remove line

packages/performance-tests/test/identity-transformer.ts
```diff
@@ -0,0 +1,79 @@
...
+import { TestIModel } from "./TestContext";
+import { Reporter } from "@itwin/perf-tools";
+import { initOutputFile, timed } from "./TestUtils";
+
+/*
+ * Tests where we perform "identity" transforms, that is just rebuilding an entire identical iModel (minus IDs)
```
MichaelBelousov, line 16: module level comments should be before the imports

packages/performance-tests/test/identity-transformer.ts
```diff
@@ -0,0 +1,79 @@
...
+ * through the transformation process.
+ */
+
+const loggerCategory = "Transformer Performance Tests Identity";
+const outputDir = path.join(__dirname, ".output");
+
+export default async function identityTransformer(iModel: TestIModel, os: any, reporter: Reporter){
```
MichaelBelousov, line 23:
```suggestion
export default async function identityTransform(iModel: TestIModel, reporter: Reporter){
```

rename and just `import * as os from "os";`, no need to take it as an argument, it's a standard import

packages/performance-tests/test/setup.ts
```diff
@@ -0,0 +1,8 @@
+import assert from "assert";
+import dotenv from "dotenv";
+
+const { error } = dotenv.config();
+
+if (error)
+  console.log("no env file found");
```
MichaelBelousov, line 7:
```suggestion
  console.error("no env file found");
```

don't you also want to throw/exit after this?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -6,6 +6,7 @@
  * @module iModels
  */
 import * as path from "path";
+import { EventEmitter } from "events";
```
MichaelBelousov, line 9: my bad, but you left in the TransformerEvent stuff I experimented with when you branched from my experimental branch. Please remove all changes to this file. You can do that with:
```sh
git checkout main packages/transformer/src/IModelTransformer.ts
```

packages/performance-tests/test/TransformerRegression.test.ts
DanRod1999: Mocha was having an issue when trying to run an `it` inside a for loop. Appears to be a known issue https://github.com/mochajs/mocha/issues/3074

packages/performance-tests/package.json
MichaelBelousov: have to update them to 4.0.999 now :) main had another PR go through :)

MichaelBelousov:
@DanRod1999 can you add a `.env.template` file? I think it was in the original branch but unless I'm missing something I don't see it here

DanRod1999:
> @DanRod1999 can you add a `.env.template` file? I think it was in the original branch but unless I'm missing something I don't see it here

Ah yeah there is one, but its named template.env. I can change the name though

MichaelBelousov:
> Ah yeah there is one, but its named template.env. I can change the name though

`template.env` is fine, don't know how I missed that

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  };
+
+  assert(process.env.OIDC_CLIENT_ID, "");
+  assert(process.env.OIDC_REDIRECT, "");
+  assert(process.env.SCOPE, "List scopes");
+  const authClient = process.env.CI === "1"
+    ? new TestBrowserAuthorizationClient({
+      clientId: process.env.OIDC_CLIENT_ID,
+      redirectUri: process.env.OIDC_REDIRECT,
+      scope: process.env.SCOPE,
```
MichaelBelousov, line 75: fuller name of `OIDC_SCOPE`?

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      authority: "https://qa-ims.bentley.com",
```
MichaelBelousov, line 76: can you use `IMJS_URL_PREFIX` env var here?

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  var reporter = new Reporter();
+  const reportPath = initOutputFile("report.csv", outputDir);
+
+  describe('Transformer Regression Tests', function () {
+    testIModels.forEach(async (value) => {
+      const iModel = value;
+        describe(`Transforms of ${iModel.name}`, async () => {
+          testCasesMap.forEach(async (testCase, key) => {
+            it(key, async () => {
+              reporter = await testCase.default(iModel, reporter);  // add timeout(0)
```
MichaelBelousov, line 114: do you need this comment?

packages/performance-tests/test/TestContext.ts
DanRod1999: I think this fixes that potential issue, but don't know if this solution is over complicated.

packages/performance-tests/test/TransformerRegression.test.ts

packages/performance-tests/test/TransformerRegression.test.ts

packages/performance-tests/test/TransformerRegression.test.ts

packages/performance-tests/package.json
DanRod1999: done

packages/performance-tests/test/TransformerRegression.test.ts

packages/performance-tests/test/TransformerRegression.test.ts
DanRod1999: I mentioned in another comment that mocha doesn't like 4 loops. Don't know how to rewrite this without a for loop, so that it could be used directly inside the mocha test instead of needing it as a middle man

packages/performance-tests/test/TransformerRegression.test.ts
MichaelBelousov: my bad, I should have thought of:

```ts
const logLevel = process.env.LOG_LEVEL
  ? Number(process.env.LOG_LEVEL)
  : LogLevel.Warning;
```

That's much less confusing

packages/performance-tests/test/TestContext.ts
```diff
@@ -0,0 +1,94 @@
...
+  
+  assert(IModelHost.authorizationClient !== undefined, "auth client undefined");
+  var briefcaseProps;
+  BriefcaseManager.getCachedBriefcases(briefcaseArg.iModelId).forEach(briefcase => {
+    if(briefcase.changeset.id === changeset.id){
+      briefcaseProps = briefcase;
+    }
+  });
+
+  if(briefcaseProps === undefined){
```
MichaelBelousov, line 73: nit: could try it this way
```ts
const briefcase = cached ?? (await ...)
```

packages/performance-tests/test/TestContext.ts
MichaelBelousov: I am resolving this in favor of new comments on the changes

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  hostConfig.authorizationClient = authClient;
+  const hubClient = new IModelsClient({ api: { baseUrl: `https://${process.env.IMJS_URL_PREFIX}api.bentley.com/imodels` } });
+  hostConfig.hubAccess = new BackendIModelsAccess(hubClient);
+  await IModelHost.startup(hostConfig);
+
+  return realizeAsyncIterator(getTestIModels());
+};
+
+// Mocha tests must know the test cases ahead time, so we collect the the Imodels first before beginning the tests
+async function realizeAsyncIterator(getTestIModels:AsyncGenerator<TestIModel>) {
```
MichaelBelousov, line 97: maybe we can rename this to `preFetchAsyncIterator`? Does that sound more understandable  than "realize" to you?

MichaelBelousov, line 97: make this function generic:

```ts
async function realizeAsyncIterator<T>(iter: AsyncIterator<T>): T[] {
  //...
```

That type is probably not exactly right, feel free to let me know if you have issues using that. Some permutation of async iterator types should accept the argument `realizeAsyncIterator(myGenerator())`

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+async function runRegressionTests() {
+  const testIModels = await setupTestData();
+  var reporter = new Reporter();
+  const reportPath = initOutputFile("report.csv", outputDir);
+
+  describe('Transformer Regression Tests', function () {
+    testIModels.forEach(async (value) => {
+      const iModel = value;
```
MichaelBelousov, line 116: is this alias really necessary? The closure passed to `it` should be able to access that value. It's only `for (let blah of)` that has this issue iirc... a `forEach` shouldn't have this problem at all, and neither should a `for (const _ of)` loop...

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  return realizeAsyncIterator(getTestIModels());
+};
+
+// Mocha tests must know the test cases ahead time, so we collect the the Imodels first before beginning the tests
+async function realizeAsyncIterator(getTestIModels:AsyncGenerator<TestIModel>) {
+  let arrImodels:TestIModel[] = [];
+  const iModelIdStr = process.env.IMODEL_IDS;
+  assert(iModelIdStr, "no Imodel Ids")
+  const iModelIds = iModelIdStr === "*" ? "" : iModelIdStr.split(",")
```
MichaelBelousov, line 101: don't handle the filtering inside this utility, it's role isn't to filter, but to flatten an existing iterator. Instead we should filter elsewhere and pass the filtered iterator to the realize function.

In this case you should probably add a `filter` callback parameter to `getIModels` so you can avoid it making unnecessary project data fetches on imodel ids that will be ignored.

packages/performance-tests/test/TransformerRegression.test.ts
DanRod1999: done

packages/performance-tests/test/TransformerRegression.test.ts
DanRod1999: done. I think i created a filter function that is sufficient, but don't know if its specifically what you intended. It is in TestUtils.ts

packages/performance-tests/test/TransformerRegression.test.ts
DanRod1999: yes this is better. Done

packages/performance-tests/test/TestContext.ts
DanRod1999: done

packages/performance-tests/test/TestContext.ts
```diff
@@ -0,0 +1,94 @@
...
+export async function downloadAndOpenBriefcase(briefcaseArg: Omit<RequestNewBriefcaseArg, "accessToken">): Promise<BriefcaseDb> {
+  const PROGRESS_FREQ_MS = 2000;
+  let nextProgressUpdate = Date.now() + PROGRESS_FREQ_MS;
+
+  const asOf = briefcaseArg.asOf ?? IModelVersion.latest().toJSON();
+  const changeset = await IModelHost.hubAccess.getChangesetFromVersion({ ...briefcaseArg, version: IModelVersion.fromJSON(asOf) });
+  
+  assert(IModelHost.authorizationClient !== undefined, "auth client undefined");
+  var briefcaseProps;
+  BriefcaseManager.getCachedBriefcases(briefcaseArg.iModelId).forEach(briefcase => {
```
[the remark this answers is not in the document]
DanRod1999: done

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  const hubClient = new IModelsClient({ api: { baseUrl: `https://${process.env.IMJS_URL_PREFIX}api.bentley.com/imodels` } });
+  hostConfig.hubAccess = new BackendIModelsAccess(hubClient);
+  await IModelHost.startup(hostConfig);
+
+  return realizeAsyncIterator(getTestIModels());
+};
+
+// Mocha tests must know the test cases ahead time, so we collect the the Imodels first before beginning the tests
+async function realizeAsyncIterator(getTestIModels:AsyncGenerator<TestIModel>) {
+  let arrImodels:TestIModel[] = [];
```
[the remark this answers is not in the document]
DanRod1999: got an error when I ran. working on it now

packages/performance-tests/test/TestContext.ts
```diff
@@ -44,13 +44,16 @@ export async function *getTestIModels() {
...
-      yield {
+      const iModelToCheck:TestIModel = {
         name: iModel.displayName,
         iModelId,
         iTwinId,
         tShirtSize: getTShirtSizeFromName(iModel.displayName),
         load: async () => downloadAndOpenBriefcase({ iModelId, iTwinId }),
-      };
+      }
+      if(filter(iModelToCheck)){
```
MichaelBelousov, line 54: make sure to run prettier + eslint, CI will fail on this spacing

packages/performance-tests/test/TransformerRegression.test.ts
MichaelBelousov: I think AsyncGenerators should be compatible with AsyncIterators but we can defer fixing that until we use this utility in more places, don't worry about it

packages/performance-tests/test/TestContext.ts
```diff
@@ -0,0 +1,89 @@
...
+  name: string;
+  iModelId: string;
+  iTwinId: string;
+  tShirtSize: string;
+  load: () => Promise<BriefcaseDb>;
+}
+
+let iTwinIdStr;
+iTwinIdStr = process.env.ITWIN_IDS;
+assert(iTwinIdStr, "no Itwins entered");
```
MichaelBelousov, line 24:
```suggestion
assert(iTwinIdStr, "no iTwins entered");
```

packages/performance-tests/test/TestContext.ts
DanRod1999: done

MichaelBelousov (approved):

packages/performance-tests/test/TransformerRegression.test.ts

PR #58, opened 2023-06-07
Don't error out on missing .env

DanRod1999:
Only error out when both .env file is missing and CI is not set

### DIFF
--- a/packages/performance-tests/test/setup.ts
+++ b/packages/performance-tests/test/setup.ts
@@ -2,5 +2,5 @@ import dotenv from "dotenv";
 
 const { error } = dotenv.config();
 
-if (error)
-  throw new Error("no env file found");
+if (error && process.env.CI !== "1")
+  throw new Error("no env file found, and not ran as a CI job");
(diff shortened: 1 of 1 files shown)


packages/performance-tests/test/setup.ts
```diff
@@ -2,5 +2,5 @@ import dotenv from "dotenv";
 
 const { error } = dotenv.config();
 
-if (error)
-  throw new Error("no env file found");
+if (error && process.env.CI !== "1")
```
MichaelBelousov, line 5: It's generally better to just check for non emptiness of an env variable than a specific value 
```ts
if (error && !process.env.CI)
```

MichaelBelousov (approved):

packages/performance-tests/test/setup.ts
DanRod1999: done. Didn't realize that would produce a boolean. good to know thanks

MichaelBelousov: it technically does but there are some caveats which I'll explain in a dm

PR #59, opened 2023-06-08
performance tests, control cases

MichaelBelousov:
- add a raw insert vs raw changeset apply (theoretical maximum) vs same content identity transform without provenance vs same content identity transform with provenance insertions test battery

### DIFF
--- a/packages/performance-tests/test/rawInserts.ts
+++ b/packages/performance-tests/test/rawInserts.ts
@@ -99,5 +99,9 @@ export default async function rawInserts(reporter: Reporter) {
       relationshipCount: IModelTransformerTestUtils.count(targetDb, "Bis.Element"),
     }
   );
+
+  sourceDb.close();
+  targetDb.close();
+  targetNoProvDb.close();
 }
 
--- a/packages/performance-tests/test/rawInserts.ts
+++ b/packages/performance-tests/test/rawInserts.ts
@@ -13,7 +13,7 @@ import { Guid, Logger, OpenMode } from "@itwin/core-bentley";
 const loggerCategory = "Raw Inserts";
 const outputDir = path.join(__dirname, ".output");
 
-const ELEM_COUNT = 1000; // 50k
+const ELEM_COUNT = 100_000;
 assert(ELEM_COUNT % 2 === 0, "elem count must be divisible by 2");
 
 export default async function rawInserts(reporter: Reporter) {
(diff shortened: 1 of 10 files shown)


packages/performance-tests/test/rawInserts.ts
```diff
@@ -0,0 +1,173 @@
...
+  const targetPath = initOutputFile(`RawInserts-Target.bim`, outputDir);
+  const targetDb = SnapshotDb.createEmpty(targetPath, { rootSubject: { name: "RawInsertsTarget" }});
+  const transformerWithProv = new IModelTransformer(sourceDb, targetDb, { noProvenance: false });
+
+  const [transformWithProvTimer] = await timed(async () => {
+    await transformerWithProv.processAll();
+  });
+
+  reporter.addEntry(
+    "Transformer Regression Tests",
```
DanRod1999, line 109: These insertion tests seem fairly different to other tests we want to create in terms of the content being tested (hence why i understand it takes place in a completely different `it`). Just curious why we choose to lump them all in the same Test Suite even though the content of what is being tested seems so different then what the others would be doing. Its not too important mechanically, but just in terms of how we organize and view this data I feel like it doesn't work well with the others tests. The way I envisioned the Test suite would be the test type that each imodel must go through, and then each test name would be the name of the iModel, so it would be very easy to look at a specific test, and see how an imodel preforms on it.

MichaelBelousov: talked on teams, agreed mostly, decided on a convention for the fields test-suite=case name, test-name=iModel name

packages/performance-tests/test/rawInserts.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  Logger.logInfo(loggerCategory, "Done. Starting with-provenance transformation of same content");
+
+  const targetPath = initOutputFile(`RawInserts-Target.bim`, outputDir);
+  const targetDb = SnapshotDb.createEmpty(targetPath, { rootSubject: { name: "RawInsertsTarget" }});
+  const transformerWithProv = new IModelTransformer(sourceDb, targetDb, { noProvenance: false });
+
+  const [transformWithProvTimer] = await timed(async () => {
+    await transformerWithProv.processAll();
+  });
```
DanRod1999, line 108: You don't log the timer like you did for identity transform. Might be useful, but not really a big deal.

MichaelBelousov: I don't think we need it for these 4 tests. Later we can add a wrapper around `reporter.addEntry` that logs the times locally and writes them to an in-progress file rather than report.csv only at the end

PR #60, opened 2023-06-09
IModelCloneContext.findTargetEntityId fixed relationship id finding

simsta6:
I had to change `Id64.isInvalid` to `!Id64.EntityReferences`, because `Id64.isInvalid` sees "e0" as valid id.
image

The problem was, that `IModelCloneContext` was trying to find relationship's target or source element that does not exist in the target iModel. So when `findTargetEntityId` returns "e0" for element referenced by relationship it would pass if statement with `Id64.isInvalid` and transformer would crash because `ECSqlStatement.bindId` can't bind invalid64Id.

Now `IModelCloneContext.findTargetEntityId` will return invalid id for target iModel relationship.


MichaelBelousov (approved):
lgtm, thanks! Dumb mistake on my part. Maybe I had to update the sourceId/targetId to EntityReferences and forgot to update the surrounding code. Stupid string types in JavaScript :/

PR #67, opened 2023-06-12
Add branch name to reporter

DanRod1999:
List branch name in reporter, so we can differentiate where the tests were ran, and compare performance based on this.
Currently I attached the branch as field in the json `info` section of the db. However, I also attached it to the imodel-name because when uploading to the db, and getting processed in power BI this should allow us to compare the run times very easily. I don't really like this solution, but I couldn't figure out a way to use the `info` fields in Power BI.

Also, I changed the logger setup, but I'm not logging the entities being processed, so I think something is wrong.

### DIFF
--- a/packages/performance-tests/test/TransformerRegression.test.ts
+++ b/packages/performance-tests/test/TransformerRegression.test.ts
@@ -39,7 +39,7 @@ const setupTestData = async () => {
   assert(LogLevel[logLevel] !== undefined, "unknown log level");
 
   Logger.initializeToConsole();
-  Logger.setLevelDefault(LogLevel.Error);
+  Logger.setLevelDefault(logLevel);
   Logger.setLevel(TransformerLoggerCategory.IModelExporter, logLevel);
   Logger.setLevel(TransformerLoggerCategory.IModelImporter, logLevel);
   Logger.setLevel(TransformerLoggerCategory.IModelTransformer, logLevel);


packages/performance-tests/test/GitUtils.ts
```diff
@@ -0,0 +1,12 @@
+import util = require("node:util");
```
MichaelBelousov, line 1: don't use `import blah = require()`, I know typescript suggests it sometimes but use either:
```ts
import fs from "fs";
//or
import * as fs from "fs";
```
whichever is fine, depends on the tsconfig settings if it's allowed.

packages/performance-tests/test/GitUtils.ts
```diff
@@ -0,0 +1,12 @@
...
+import childPoc = require("node:child_process");
```
MichaelBelousov, line 2: in iTwin.js JavaScript we prefer not to use the `node:` prefix in imports. Even if I agree it's better, it's inconsistent with our other code

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -33,15 +34,15 @@ const testCasesMap = new Map([
...
+  const logLevel = process.env.LOG_LEVEL ? Number(process.env.LOG_LEVEL) : LogLevel.Error;
 
   assert(LogLevel[logLevel] !== undefined, "unknown log level");
 
   Logger.initializeToConsole();
-  Logger.setLevelDefault(logLevel);
-  Logger.setLevel(TransformerLoggerCategory.IModelExporter, LogLevel.Warning);
-  Logger.setLevel(TransformerLoggerCategory.IModelImporter, LogLevel.Warning);
-  Logger.setLevel(TransformerLoggerCategory.IModelTransformer, LogLevel.Warning);
+  Logger.setLevelDefault(LogLevel.Error);
```
MichaelBelousov, line 42: the LOG_LEVEL env variable should affect the default log level, not the log level of the transformer only. Add another variable if you want to configure that separately.

packages/performance-tests/test/rawInserts.ts
```diff
@@ -18,7 +18,7 @@ const iModelName = "Many PhysicalObjects and Relationships";
 const ELEM_COUNT = 100_000;
```
MichaelBelousov, line 18: right above this (github won't let me comment on it) is `iModelName`. Don't you want to append the branch name to that?

packages/performance-tests/test/cases/identity-transformer.ts
```diff
@@ -66,11 +66,12 @@ export default async function identityTransformer(iModel: TestIModel, reporter:
       "Id": iModel.iModelId,
       "T-shirt size": iModel.tShirtSize,
       "Gb size": sizeInGb,
+      "Branch Name": branchName,
       /* eslint-enable @typescript-eslint/naming-convention */
     };
     reporter.addEntry(
       "identity transform (provenance)",
-      iModel.name,
+      `${iModel.name} - ${branchName}`,
```
MichaelBelousov, line 74: can `branchName` come first? Feels a bit less ambiguous to prefix rather than suffix, I imagine there are many iModels using hyphens. Perhaps we should prefer a colon. For example:

`federation-guid-optiization: Highway Restoration XZT-309 - Display Model`

packages/performance-tests/test/rawInserts.ts
DanRod1999: don't have access to branch branchName here so adding it to the reporter entries

packages/performance-tests/test/GitUtils.ts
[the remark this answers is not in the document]
DanRod1999: tried this but then couldn't promisfy exec unless I changed the name of the const exec here
`const exec = util.promisify(child_proc.exec);`

Don't know if there's a better what to write this

packages/performance-tests/test/GitUtils.ts
DanRod1999: done

packages/performance-tests/test/GitUtils.ts
DanRod1999: done

packages/performance-tests/test/cases/identity-transformer.ts
DanRod1999: done

packages/performance-tests/test/TransformerRegression.test.ts

packages/performance-tests/test/GitUtils.ts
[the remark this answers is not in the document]
MichaelBelousov: fair, resolving

MichaelBelousov (approved):

Issue #61, opened 2023-06-09
Foreign Key Constraint error when deleting an element that is referenced in other elements as code scope


DeividasDavidav:
**Describe the bug**
When using onDeleteElement() function from IModelImporter to delete an element, if an element is referenced by other elements in their code scopes, it throws Foreign key constraint error. That is because ElementTreeDeleter doesn't find elements with code scope references to specific element when deleting it. (Already created an issue with possible solution in itwins-core (https://github.com/iTwin/itwinjs-core/issues/5576) but they said that ElementTreeDeleter shouldn't solve this relationship issue)

**To Reproduce**
1. Create the first element
2. Create the second element that references the first element in its code scope
3. Try to delete the first element (can use onDeleteElement() from IModelImporter)
4. See "error deleting element" error (errorNumber: 65553)

**Expected behavior**
When deleting an element, it should also process the elements that reference the element in their code scopes.

**Desktop (please complete the applicable information):**

- iTwin.js Version 3.6.1
- Node Version 18.16.0

MichaelBelousov:
We can port your suggested fix to a `ElementCascadingDeleter` in the transformer. Given you already have sample fix code in your original core issue, would you be willing to make a PR?

Do we need to add additional sanity checks though that the transformer should only delete this element if its other references (the scope one that currently errors) are also being deleted?

MichaelBelousov:
this was fixed by #75, closing

PR #75, opened 2023-06-16
Add ElementCascadingDeleter to fix FK errors while deleting element which is referenced in code scopes

DeividasDavidav:
Added ElementCascadingDeleter, which works similarly to ElementTreeDeleter, but also finds elements that reference this element in their code scopes and deletes those elements.

https://github.com/iTwin/imodel-transformer/issues/61


packages/transformer/src/ElementCascadingDeleter.ts
```diff
@@ -0,0 +1,46 @@
...
+/** Deletes an entire element tree, including sub-models, child elements and code scope references.
+  * Items are deleted in bottom-up order. Definitions and Subjects are deleted after normal elements.
+  * Call deleteNormalElements on each tree. Then call deleteSpecialElements.
+  */
+export class ElementCascadingDeleter extends ElementTreeDeleter {
+  protected shouldVisitCodeScopes(_elementId: Id64String, _scope: ElementTreeWalkerScope) { return true; }
+
+  /** The main tree-walking function */
+  protected override processElementTree(element: Id64String, scope: ElementTreeWalkerScope): void {
+    if(this.shouldVisitCodeScopes(element, scope)) {
```
MichaelBelousov, line 30: resolve at will, why is the linter/prettier-check not configured to reject this?
```suggestion
    if (this.shouldVisitCodeScopes(element, scope)) {
```

packages/transformer/src/ElementCascadingDeleter.ts
```diff
@@ -0,0 +1,46 @@
...
+      this._processCodeScopes(element, scope);
+    }
+    super.processElementTree(element, scope);
+  }
+  /** Process code scope references */
+  private _processCodeScopes(element: Id64String, scope: ElementTreeWalkerScope) {
+    const newScope = new ElementTreeWalkerScope(scope, element);
+    this._iModel.withPreparedStatement("select ECInstanceId from bis.Element where CodeScope.id=? and Parent.id is null", (stmt) => {
```
MichaelBelousov, line 38: nit: prefer spacing sql over multiple lines and capitalizing keywords to be consistent with this codebase:
```suggestion
    this._iModel.withPreparedStatement(`
      SELECT ECInstanceId
      FROM bis.Element
      WHERE CodeScope.Id=?
        AND Parent.Id IS NULL
    `, (stmt) => {
```

packages/transformer/src/ElementCascadingDeleter.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @module iModels
+ */
+import { ElementTreeDeleter, ElementTreeWalkerScope, IModelDb } from "@itwin/core-backend";
+import { DbResult, Id64String } from "@itwin/core-bentley";
+
+/** Deletes an element tree and code scope references starting with the specified top element. The top element is also deleted. Uses ElementCascadeDeleter.
+ * @param iModel The iModel
+ * @param topElement The parent of the sub-tree
+ */
+export function deleteElementCascadeTree(iModel: IModelDb, topElement: Id64String): void {
```
MichaelBelousov, line 15: How about `deleteElementTreeCascade`? Sounds slightly clearer.

MichaelBelousov (dismissed):
LGTM, minor comments

Issue #68, opened 2023-06-14
[fed guid optimization] Change processing does not work in reverse synchronization workflow


ViliusRuskys:
**Describe the bug**
The issue is related to the new rule added to the federation guid branch of imodel-transformer. This rule enforces that for change processing workflow the source iModel starting changeset index must be 1 higher over the changeset index used for the previous transformation.
Right now we crate a Forked iModel by running the transformation in normal direction (Lets say source only has 1 changeset)

**S1** -> **T1**, **T2**, **T3** _(During the initial transformation target iModel gets multiple changes pushed so target (T) is on changeset 3)_
&emsp;&emsp;&ensp;**T4** _(user pushes a few more changes to forked iModel)_
**S2** <- **T5** _(we merge back with reverse synchronization workflow using start changeset with index 4 because data flow is reversed)_
The transformation errors out, because transformer expects changeset with index 2 to be used and isntead it got index 4.

Comparing these changesets does not make sense, because these changesets are from different iModels as well (changeset 1 is in source, changeset 4 is in target).

MichaelBelousov:
@ViliusRuskys we may need to discuss this one more.
I have added to the "big branch workflow test" a [custom synchronizer that always performs multi-changeset synchronizations](https://github.com/iTwin/imodel-transformer/pull/77). But the tests are not failing with the error you say they would here. Is there anything in that implementation that looks wrong to you?

Here is the resulting change history of the test after those changes:

```
┌─────────┬───────────┬───────────┬───────────┐
│ (index) │  master   │  branch1  │  branch2  │
├─────────┼───────────┼───────────┼───────────┤
│    0    │   '0 '    │           │           │
│    1    │   '0 '    │ '1 2d730' │           │
│    2    │ '1 66678' │ '1 2d730' │           │
│    3    │ '1 66678' │ '1 2d730' │ '1 6f3c0' │
│    4    │ '1 66678' │ '2 d6b09' │ '1 6f3c0' │
│    5    │ '1 66678' │ '3 96a20' │ '1 6f3c0' │
│    6    │ '1 66678' │ '4 d25c9' │ '1 6f3c0' │
│    7    │ '1 66678' │ '5 36512' │ '1 6f3c0' │
│    8    │ '1 66678' │ '6 5101d' │ '1 6f3c0' │
│    9    │ '3 c6100' │ '6 5101d' │ '1 6f3c0' │
[...]

PR #78, opened 2023-06-21
reverse synchronization test case

ViliusRuskys:
Test case to reproduce issue #68


MichaelBelousov:
this example skips the initialize fork changeset which is not legal in the new strict startChangeset rules. Can you reproduce else how while using the correct startChangeset?.

Note that while we intend to keep the strictness, we have not yet mitigated a performance penalty for including that changeset which is recorded in #84

MichaelBelousov:
found the bug, first synchronization did not correctly set reverseSync version

PR #80, opened 2023-06-22
Add pending reference resolution when referenced element is not exported

mindaugasdirg:
If element is not exported all pending references are left in memory until transformation completes. This causes memory issues when a lot of elements are not exported. To reduce memory usage transformer can resolve pending references as soon as it knows that element won't be exported.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Use Node.js 18
       uses: actions/setup-node@v3
       with:
-        node-version: 18
+        node-version: 18.16.0
         cache: 'pnpm'
 
     - name: Install dependencies
@@ -51,7 +51,7 @@ jobs:
     - name: Use Node.js 18
       uses: actions/setup-node@v3
       with:
-        node-version: 18
+        node-version: 18.16.0
         cache: 'pnpm'
 
     - id: set-matrix
@@ -107,7 +107,7 @@ jobs:
     - name: Use Node.js 18
       uses: actions/setup-node@v3
       with:
-        node-version: 18
+        node-version: 18.16.0
         cache: 'pnpm'
 
     - name: Force dependency resolution
--- a/change/@itwin-imodel-transformer-af619951-5fbf-47a8-a92c-7c1027ff2cb1.json
+++ b/change/@itwin-imodel-transformer-af619951-5fbf-47a8-a92c-7c1027ff2cb1.json
@@ -1,5 +1,5 @@
 {
-  "type": "minor",
+  "type": "patch",
   "comment": "Add heap memory optimizations",
   "packageName": "@itwin/imodel-transformer",
   "email": "Mindaugas.Dirgincius@bentley.com",
--- a/change/@itwin-imodel-transformer-af619951-5fbf-47a8-a92c-7c1027ff2cb1.json
+++ b/change/@itwin-imodel-transformer-af619951-5fbf-47a8-a92c-7c1027ff2cb1.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Add heap memory optimizations",
+  "comment": "Add pending reference resolution when referenced element is not exported",
   "packageName": "@itwin/imodel-transformer",
   "email": "Mindaugas.Dirgincius@bentley.com",
   "dependentChangeType": "patch"
--- a/change/@itwin-imodel-transformer-af619951-5fbf-47a8-a92c-7c1027ff2cb1.json
+++ b/change/@itwin-imodel-transformer-af619951-5fbf-47a8-a92c-7c1027ff2cb1.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add heap memory optimizations",
+  "packageName": "@itwin/imodel-transformer",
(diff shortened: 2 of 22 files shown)


MichaelBelousov:
is this still WIP as marked in the title? If so can you mark it as a draft?

mindaugasdirg:
Sorry, forgot to remove WIP from the title. PR is ready for review.

mindaugasdirg:
I've also updated pipeline since tests were facing SegFault issue mentioned here: https://github.com/iTwin/itwinjs-core/issues/5663#issuecomment-1600764489

packages/transformer/src/IModelTransformer.ts
```diff
@@ -755,6 +772,27 @@ export class IModelTransformer extends IModelExportHandler {
...
+  public override onSkipElement(sourceElement: Element): void {
+    if (this.context.findTargetElementId(sourceElement.id) !== Id64.invalid) {
+      // element already has provenance
+      return;
+    }
+
+    Logger.logInfo(loggerCategory, `Element '${sourceElement.id}' won't be exported. Marking its references as resolved`);
+    // Remap skipped element to non-existent EcInstanceId.
+    // When exporting other elements transformer needs reset all references to missingReferenceId
+    this.context.remapElement(sourceElement.id, missingReferenceId);
```
MichaelBelousov, line 784: This is a bit too hacky... and imo not future proof. Not likely but that 0x2 id could eventually be taken or elsehow made invalid for usage. We really need to move that native element remap table into TypeScript. I have some initial work for it but because of how geometry is processed in the native layer, it's not simple, so I haven't finished it.

I think it's possible to instead for now introduce a set of missing reference ids that must be checked in onTransformBlah and maybe before creating pending references. Can you try that instead?

packages/transformer/src/IModelExporter.ts
```diff
@@ -82,6 +82,9 @@ export abstract class IModelExportHandler {
    */
   public shouldExportElement(_element: Element): boolean { return true; }
 
+  /** Called when element is skipped instead of exported. */
+  public onSkipElement(_element: Element): void { }
```
MichaelBelousov, line 86: can we make this new API take a parameter of `{ id: Id64String }` type? I didn't do a full check but on the surface the code that uses this looks like maybe it can take just an id. I'd like to eventually make it possible to not load the element if `shouldExportElement` returned false, and this API taking an element would violate that.

MichaelBelousov (changes requested):

.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     - name: Use Node.js 18
       uses: actions/setup-node@v3
       with:
-        node-version: 18
+        node-version: 18.16.0
```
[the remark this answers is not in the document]
MichaelBelousov: you put in a comment about this apparently, thanks, resolving

packages/transformer/src/IModelTransformer.ts
mindaugasdirg: Added skipped entities property for tracking entities that are not exported. Although for now made it only track elements since it should bring the most benefit as most relationships are between elements. Also solution seems cleaner since actually checks onTransform___ are no longer needed since current onTransform implementations already handle cases when reference is missing. The only thing that was needed to be added was additonal check so that pending references wouldn't be added to already skipped elements.

packages/transformer/src/IModelExporter.ts
mindaugasdirg: I thought about only passing id at first, but saw that to get pending references to skipped element required all entity. I've changed onSkipElement to accept id directly, but had to also add `getReferencersByEntityKey` method to `PendingReferenceMap` class.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -609,6 +610,21 @@ export class IModelTransformer extends IModelExportHandler {
     Logger.logWarning(loggerCategory, `Tried to defer/skip an element, which is no longer necessary`);
   }
 
+  private isRelatedElementProps(obj: any): obj is RelatedElementProps {
+    return typeof (obj) === "object" && "id" in obj;
```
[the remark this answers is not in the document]
mindaugasdirg: I removed the method since it's no longer needed after your suggestion with skipped entity set.

MichaelBelousov:
thinking about this some more, instead of keeping the pending references in memory and only updating them, can we just keep the id of the incomplete element, and after all have been inserted, re-transform all incomplete elements from scratch? So same as what we do now, but rather than update elements as soon as we find their references, wait until the end when everything is inserted, and batch update them.

change/@itwin-imodel-transformer-af619951-5fbf-47a8-a92c-7c1027ff2cb1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
MichaelBelousov, line 2: we're introducing a new IModelExportHandler API, can we make this a minor version bump?

MichaelBelousov (approved):
LGTM, I'll add a separate issue about fixing unbounded cache sizes which includes the two-pass reference resolution that I mentioned

MichaelBelousov (approved):

PR #81, opened 2023-06-22
Dan/perf regression tests

DanRod1999:
Add prepare-fork tests, and support testing on imodels generated with minimal or all fed-guids on elements


packages/performance-tests/test/iModelUtils.ts
```diff
@@ -0,0 +1,82 @@
...
+  nativeDb.setITwinId(Guid.empty); // empty iTwinId means "standalone"
+  nativeDb.saveChanges(); // save change to iTwinId
+  nativeDb.deleteAllTxns(); // necessary before resetting briefcaseId
+  nativeDb.resetBriefcaseId(BriefcaseIdValue.Unassigned); // standalone iModels should always have BriefcaseId unassigned
+  nativeDb.saveLocalValue("StandaloneEdit", JSON.stringify({ txns: true }));
+  nativeDb.saveChanges(); // save change to briefcaseId
+  nativeDb.closeIModel();
+}
+
+export function generateTestIModel(iModelParam: IModelParams): TestIModel {
```
DanRod1999, line 32: Should I just use this function in the rawInserts test, and just set a timer on it, so we don't need essentially the same code in 2 places? I don't believe there's many differences. I think maybe just adding a file name into that iModelParams interface

packages/performance-tests/test/TestContext.ts
```diff
@@ -8,9 +8,33 @@ import { Logger } from "@itwin/core-bentley";
 import { BriefcaseIdValue, IModelVersion } from "@itwin/core-common";
 import { AccessTokenAdapter, BackendIModelsAccess } from "@itwin/imodels-access-backend";
 import assert from "assert";
+import { generateTestIModel, IModelParams } from "./iModelUtils";
 
 const loggerCategory = "TestContext";
 
+export interface ReporterInfo {
```
DanRod1999, line 15: Is this a good place for these helper interfaces. Or maybe move them into a reporterUtils file? Don't know if its better to add to this file, or create yet another helper file. Also i've been considering moving those util files into a `RegTestUtils` dir or something

packages/performance-tests/test/cases/identity-transformer.ts
```diff
@@ -61,24 +64,15 @@ export default async function identityTransformer(iModel: TestIModel, reporter:
...
     };
-    reporter.addEntry(
-      "identity transform (provenance)",
-      `${branchName}: ${iModel.name}`,
-      "time elapsed (seconds)",
-      entityProcessingTimer?.elapsedSeconds ?? -1,
-      record
-    );
     targetDb.close();
     sourceDb.close();
```
MichaelBelousov, line 74: so we should probably never mutate the source db in test cases and not close it until after all test cases for that iModel have finished. Let's not close it here.

packages/performance-tests/test/cases/prepare-fork.ts
```diff
@@ -0,0 +1,92 @@
...
+  const filePath = sourceDb.pathName;
+  fs.copyFileSync(filePath, branchPath);
+  setToStandalone(branchPath);
+  const branchDb = StandaloneDb.openFile(branchPath);
+
+  let entityProcessingTimer: StopWatch | undefined;
+  try {
+    [entityProcessingTimer] = await timed(async () => {
+      // create an external source and owning repository link to use as our *Target Scope Element* for future synchronizations
+      const masterLinkRepoId = new RepositoryLink({
```
MichaelBelousov, line 37: move most of this callback into its own function, maybe called `classicalTransformerBranchInit`

packages/performance-tests/test/cases/prepare-fork.ts
```diff
@@ -0,0 +1,92 @@
...
+      await branchInitializer.processAll();
+      // save+push our changes to whatever hub we're using
+      const description = "initialized branch iModel";
+      branchDb.saveChanges(description);
+
+      branchDb.close();
+      sourceDb.close();
+      branchInitializer.dispose();
+    });
+    Logger.logInfo(loggerCategory, `Preparefork time: ${entityProcessingTimer.elapsedSeconds}`);
```
MichaelBelousov, line 75:
```suggestion
    Logger.logInfo(loggerCategory, `Prepare Fork time: ${entityProcessingTimer.elapsedSeconds}`);
```

packages/performance-tests/test/iModelUtils.ts
MichaelBelousov: sounds ideal. Not sure where you need a timer though?

packages/performance-tests/test/cases/identity-transformer.ts
MichaelBelousov: let's also try to make sure we open the sourceDb as readonly, the only case I can think of where we may want it writeable is the init fork test where we add our own fed guids to an empty imodel before initialization, and in that one, the test case can manually copy the underlying sqlite file and open the new one as writeable, then discard it at the end of the test.

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -95,16 +99,55 @@ const setupTestData = async () => {
...
+            "Id": iModel.iModelId,
+            "T-shirt size": iModel.tShirtSize,
+            "Gb size": sizeInGb,
+            "Branch Name": branchName,
+            "Federation Guid Saturation": fedGuidSaturation,
+            /* eslint-enable @typescript-eslint/naming-convention */
+          };
+          sourceBriefcaseArgs = {
+            fileName: sourceDb.pathName,
+            briefcaseId: sourceDb.briefcaseId,
```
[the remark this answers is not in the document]
DanRod1999: I was basing this off of how we open a file in `downloadAndOpenBriefcase`, so I just assumed a briefcaseId was necessary for `BriefcaseDb.open` but I see now that it is used to control read only or not.

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -95,16 +99,55 @@ const setupTestData = async () => {
...
+          };
+        });
+
         testCasesMap.forEach(async (testCase, key) => {
           it(key, async () => {
-            reporter = await testCase(iModel, reporter, branchName);
+            const reporterEntry = await testCase(sourceDb, sourceBriefcaseArgs);
```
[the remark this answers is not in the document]
DanRod1999: Basically you're saying what if we want to report multiple entries from a given test? I didn't think of this use case. I just thought it simplified the tests to not pass around this reporter and to also to have one consistent place where entries are added. If you think its better to keep it within each test I'm fine with that. The only other solution I have would be to return a list of `ReporterEntry` objects and have a forloop or map that performs an addEntry on each item in the list of reports from each test. I don't particularly favor one over the other, so whatever you think is best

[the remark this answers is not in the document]
MichaelBelousov: I think it's most flexible to have the test cases receive a reporter instance as an argument. If the amount of parameters they need passes 3, they can instead receive a single object parameter containing a reporter and everything else.

packages/performance-tests/test/TestContext.ts
DanRod1999: done

packages/performance-tests/test/TestContext.ts
```diff
@@ -8,9 +8,33 @@ import { Logger } from "@itwin/core-bentley";
...
+  /* eslint-disable @typescript-eslint/naming-convention */
+  "Id": string;
+  "T-shirt size": string;
+  "Gb size": string;
+  "Branch Name": string;
+  "Federation Guid Saturation": number;
```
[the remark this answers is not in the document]
DanRod1999: done

packages/performance-tests/test/TestContext.ts
```diff
@@ -56,6 +80,16 @@ export async function *getTestIModels(filter: (iModel: TestIModel) => boolean) {
       }
     }
   }
+  const iModelParamsFedguids: IModelParams = {
```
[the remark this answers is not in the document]
DanRod1999: done

packages/performance-tests/test/TransformerRegression.test.ts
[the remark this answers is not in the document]
MichaelBelousov: yeah, let's use a pull-only briefcase and if we can remove the BriefcaseArgs type

packages/performance-tests/test/cases/identity-transformer.ts
```diff
@@ -9,24 +9,27 @@
...
-  const targetDb = SnapshotDb.createEmpty(targetPath, {rootSubject: {name: iModel.name}});
+export default async function identityTransformer(sourceDb: BriefcaseDb, sourceBriefcaseArgs: BriefcaseArgs) {
+
+  if(!sourceDb.isOpen)
+    sourceDb = await BriefcaseDb.open({
+      fileName: sourceBriefcaseArgs.fileName,
+      readonly: sourceBriefcaseArgs.briefcaseId ? sourceBriefcaseArgs.briefcaseId === BriefcaseIdValue.Unassigned : false,
+    });
+
+  const targetPath = initOutputFile(`${sourceDb.iTwinId}-${sourceDb.name}-target.bim`, outputDir);
```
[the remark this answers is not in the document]
DanRod1999: done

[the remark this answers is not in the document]
DanRod1999: done

packages/performance-tests/test/cases/identity-transformer.ts
DanRod1999: so just to clarify, we want to move open and close into the base test file, and so something like this?
`
testCasesMap.forEach(async (testCase, key) => {...
  sourceDb.open()//readonly
  it(key, async () => {
    reporter = await testCase(sourceDb, sourceBriefcaseArgs, reporter, record);
  }).timeout(0);
  sourceDb.close()
`

MichaelBelousov: yes basically, and make sure to add a comment on the `close()` saying that you're doing it to ensure connection cache reusage doesn't affect results.

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -95,16 +99,55 @@ const setupTestData = async () => {
...
   const reportPath = initOutputFile("report.csv", outputDir);
   const branchName =  await getBranchName();
 
   describe("Transformer Regression Tests", function () {
     testIModels.forEach(async (iModel) => {
       describe(`Transforms of ${iModel.name}`, async () => {
+        let sourceDb: BriefcaseDb;
+        let record: ReporterInfo;
+        let sourceBriefcaseArgs: BriefcaseArgs;
+        before( async () => {
```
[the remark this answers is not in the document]

packages/performance-tests/test/cases/identity-transformer.ts
DanRod1999: done

packages/performance-tests/test/TransformerRegression.test.ts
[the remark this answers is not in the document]

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -95,16 +99,55 @@ const setupTestData = async () => {
...
+          Logger.logInfo(loggerCategory, `processing iModel '${iModel.name}' of size '${iModel.tShirtSize.toUpperCase()}'`);
+          sourceDb = await iModel.load();
+          const fedGuidSaturation = sourceDb.withStatement(
+            "SELECT CAST(SUM(hasGuid) as DOUBLE)/SUM(total) ratio FROM (SELECT IIF(FederationGuid IS NOT NULL, 1, 0) hasGuid, 1 as total FROM bis.Element)",
```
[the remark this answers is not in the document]

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -95,16 +99,55 @@ const setupTestData = async () => {
...
+            (stmt) => {
+              stmt.step();
```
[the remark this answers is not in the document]

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -95,16 +99,55 @@ const setupTestData = async () => {
...
+              return stmt.getValue(0).getDouble();
+            }
+          );
+          Logger.logInfo(loggerCategory, `Federation Guid Saturation '${fedGuidSaturation}'`);
+          const toGb = (bytes: number) => `${(bytes / 1024 **3).toFixed(2)}Gb`;
+          const sizeInGb = toGb(fs.statSync(sourceDb.pathName).size);
+          Logger.logInfo(loggerCategory, `loaded (${sizeInGb})'`);
+          record = {
+            /* eslint-disable @typescript-eslint/naming-convention */
```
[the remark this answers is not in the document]
DanRod1999: had issues still working on this

packages/performance-tests/test/cases/identity-transformer.ts
```diff
@@ -9,24 +9,27 @@
...
-  Logger.logInfo(loggerCategory, `processing iModel '${iModel.name}' of size '${iModel.tShirtSize.toUpperCase()}'`);
-  const sourceDb = await iModel.load();
-  const toGb = (bytes: number) => `${(bytes / 1024 **3).toFixed(2)}Gb`;
-  const sizeInGb = toGb(fs.statSync(sourceDb.pathName).size);
-  Logger.logInfo(loggerCategory, `loaded (${sizeInGb})'`);
-  const targetPath = initOutputFile(`${iModel.iTwinId}-${iModel.name}-target.bim`, outputDir);
-  const targetDb = SnapshotDb.createEmpty(targetPath, {rootSubject: {name: iModel.name}});
+export default async function identityTransformer(sourceDb: BriefcaseDb, sourceBriefcaseArgs: BriefcaseArgs) {
+
+  if(!sourceDb.isOpen)
```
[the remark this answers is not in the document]
DanRod1999: done

packages/performance-tests/test/TransformerRegression.test.ts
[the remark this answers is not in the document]

packages/performance-tests/test/TestContext.ts
```diff
@@ -8,9 +8,15 @@ import { Logger } from "@itwin/core-bentley";
 import { BriefcaseIdValue, IModelVersion } from "@itwin/core-common";
 import { AccessTokenAdapter, BackendIModelsAccess } from "@itwin/imodels-access-backend";
 import assert from "assert";
+import { generateTestIModel } from "./iModelUtils";
 
 const loggerCategory = "TestContext";
 
+export interface BriefcaseArgs {
```
MichaelBelousov, line 15: do we still need this?

packages/performance-tests/test/iModelUtils.ts
DanRod1999: looks like there's a issue with the return type here for `generateTestIModel`, so maybe I can't do this, or its least not as simply. I didn't realize the sourceDb created from this gets used throughout the rest of the insert tests

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -102,17 +107,60 @@ async function runRegressionTests() {
...
+        before(async () => {
+          Logger.logInfo(loggerCategory, `processing iModel '${iModel.name}' of size '${iModel.tShirtSize.toUpperCase()}'`);
+          sourceDb = await iModel.load();
+          const fedGuidSaturation = sourceDb.withStatement(
+            `
+            SELECT
+            CAST(SUM(hasGuid) as DOUBLE)/SUM(total) ratio 
+            FROM (
+              SELECT IIF(FederationGuid IS NOT NULL, 1, 0) hasGuid,
+              1 as total FROM bis.Element
```
MichaelBelousov, line 122:
```suggestion
              1 AS total FROM bis.Element
```

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -102,17 +107,60 @@ async function runRegressionTests() {
...
+
+        before(async () => {
+          Logger.logInfo(loggerCategory, `processing iModel '${iModel.name}' of size '${iModel.tShirtSize.toUpperCase()}'`);
+          sourceDb = await iModel.load();
+          const fedGuidSaturation = sourceDb.withStatement(
+            `
+            SELECT
+            CAST(SUM(hasGuid) as DOUBLE)/SUM(total) ratio 
+            FROM (
+              SELECT IIF(FederationGuid IS NOT NULL, 1, 0) hasGuid,
```
MichaelBelousov, line 121:
```suggestion
              SELECT IIF(FederationGuid IS NOT NULL, 1, 0) AS hasGuid,
```

MichaelBelousov: nit, feel free to resolve as not going to do

packages/performance-tests/test/cases/prepare-fork.ts
DanRod1999: which part should be put into a function, everything inside of the timer? Otherwise, I don't know how to put the setup in a function when we are currently timing some of that process.

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -102,17 +107,60 @@ async function runRegressionTests() {
...
+        let sourceDb: BriefcaseDb;
+        let reportInfo: ReporterInfo
+
+        before(async () => {
+          Logger.logInfo(loggerCategory, `processing iModel '${iModel.name}' of size '${iModel.tShirtSize.toUpperCase()}'`);
+          sourceDb = await iModel.load();
+          const fedGuidSaturation = sourceDb.withStatement(
+            `
+            SELECT
+            CAST(SUM(hasGuid) as DOUBLE)/SUM(total) ratio
```
MichaelBelousov, line 119: instead of `SUM(total)` can't you just do `COUNT(*)` and remove the `total` column from the query?

packages/performance-tests/test/TransformerRegression.test.ts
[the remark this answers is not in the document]

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -102,17 +107,60 @@ async function runRegressionTests() {
...
+            "T-shirt size": iModel.tShirtSize,
+            "Gb size": sizeInGb,
+            "Branch Name": branchName,
+            "Federation Guid Saturation 0-1": fedGuidSaturation,
+          };
+        });
+
         testCasesMap.forEach(async (testCase, key) => {
+          before(async () => {
+            if(!sourceDb.isOpen){
```
MichaelBelousov, line 144: still not sure why we need this code? We know that we close it, we should always have to reopen it. I'd prefer we replace the `testIModel.load()` function with like `testIModel.getFilePath()` and just always open+close a fresh one using that, never loading it like we used to.

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -102,17 +107,60 @@ async function runRegressionTests() {
...
+              });
+            }
+          });
+
           it(key, async () => {
-            reporter = await testCase(iModel, reporter, branchName);
+            reporter = await testCase(sourceDb, reporter, reportInfo);
           }).timeout(0);
+          
+          sourceDb.close();
```
MichaelBelousov: this can execute before the async callback passed to `it(key, ...)` is called. You should close the sourceDb inside the async callback body of `it(key...)`

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -102,17 +107,60 @@ async function runRegressionTests() {
...
         });
       });
     });
 
     const _15minutes = 15 * 60 * 1000;
 
-    it("Transform vs raw inserts", async () => {
+    it.skip("Transform vs raw inserts", async () => {
```
MichaelBelousov, line 163: don't leave this in the PR, if the PR is not ready for review, draft it

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -102,17 +107,60 @@ async function runRegressionTests() {
...
+              sourceDb = await BriefcaseDb.open({
+                fileName: sourceDb.pathName,
+                readonly: true,
+              });
+            }
+          });
+
           it(key, async () => {
-            reporter = await testCase(iModel, reporter, branchName);
+            reporter = await testCase(sourceDb, reporter, reportInfo);
```
MichaelBelousov, line 153: is it worth instead doing like:

```ts
await testCase(sourceDb, {
  addReport: (smallReportSubset) => {
    reporter.addEntry( ...smallReportSubset, additionalInfo: reportInfo });
  }
);
```

This removes the two way relationship where the consumer must correctly send `reportInfo` into the `reporter`

packages/performance-tests/test/cases/prepare-fork.ts
MichaelBelousov: yes basically all of it

packages/performance-tests/test/iModelUtils.ts
MichaelBelousov: refactor this to return an `IModelDb` and wrap the `IModelDb` in a `TestModel` in the caller that needs it

packages/performance-tests/test/iModelUtils.ts
```diff
@@ -0,0 +1,82 @@
...
+  let sourceDb = StandaloneDb.createEmpty(sourcePath, { rootSubject: { name: `testIModel-fedguids-${iModelParam.fedGuids}.bim` }});
+  const pathName = sourceDb.pathName;
+  sourceDb.close();
+  setToStandalone(pathName);
+  sourceDb = StandaloneDb.openFile(sourcePath, OpenMode.ReadWrite);
+
+  const physModelId = PhysicalModel.insert(sourceDb, IModelDb.rootSubjectId, "physical model");
+  const categoryId = SpatialCategory.insert(sourceDb, IModelDb.dictionaryId, "spatial category", {});
+
+  // 100,000 elements, 50,000  relationships
```
MichaelBelousov, line 46: can you remove this comment? It's not accurate since the caller controls it

packages/performance-tests/test/iModelUtils.ts
```diff
@@ -0,0 +1,82 @@
...
+      category: categoryId,
+      geom: IModelTransformerTestUtils.createBox(Point3d.create(i, i, i)),
+      placement: {
+        origin: Point3d.create(i, i, i),
+        angles: YawPitchRollAngles.createDegrees(i, i, i),
+      },
+      model: physModelId,
+      code: new Code({ spec: IModelDb.rootSubjectId, scope: IModelDb.rootSubjectId, value: `${2*i + n}`}),
+      userLabel: `${2*i + n}`,
+      federationGuid: iModelParam.fedGuids ? Guid.createValue() : Guid.empty, // Guid.empty = 00000000-0000-0000-0000-000000000000
```
MichaelBelousov, line 59: prefer to leave undefined over createValue, it's much more efficient to create them in native which will happen automatically if it's undefined

packages/performance-tests/test/TransformerRegression.test.ts
DanRod1999: I had this if because on the first test the db would always be open. So then I didn't want to call open again if it already was. But removing load would fix this problem, so I'll do that

packages/performance-tests/test/TransformerRegression.test.ts
DanRod1999: Not sure I fully understand what makes this a two way relationship? However, I needed `reportInfo` in order to get the branch name which is needed for the test name. If this still gives access to reportInfo to use in smallReportSubset, then this is a change I could make, but I guess I don't understand the benefit than?

packages/performance-tests/test/cases/prepare-fork.ts
DanRod1999: leaving out the duplicating part that we weren't timing, but if you want that included I will add. Otherwise done

packages/performance-tests/test/TestContext.ts
DanRod1999: done

packages/performance-tests/test/TransformerRegression.test.ts

packages/performance-tests/test/TransformerRegression.test.ts

packages/performance-tests/test/TransformerRegression.test.ts

packages/performance-tests/test/TransformerRegression.test.ts
MichaelBelousov: Can't we generate the test name here anyway? We know the iModel, test case, and branch here. This function could add a parameter for tests to add a subname if they want to add multiple entries

packages/performance-tests/test/TransformerRegression.test.ts
DanRod1999: this seemed to cause the same error when trying to use an imodel that is closed, but that may be due to issues with the if stmt inside the before, so hopefully fixing that will also fix this.

packages/performance-tests/test/TransformerRegression.test.ts
DanRod1999: Having trouble with this. I don't fully understand the getFilePath() function's purpose. I've tried something else instead that I think works, and when debugging everything looks in order and I can collect the FedGuid saturation and imodel size. However I think its trying to run the tests before first running the code inside the `before()` stmt, so the sourcedb is never initialized. I could be wrong, but when debugging, a breakpoint in line 154 gets reached before 122.

DanRod1999: and the error is `TypeError: Cannot read properties of undefined (reading 'getFilePath')`, caused by ln 157 of this file

packages/performance-tests/test/iModelUtils.ts
DanRod1999: done

packages/performance-tests/test/iModelUtils.ts
DanRod1999: done

packages/performance-tests/test/cases/prepare-fork.ts
DanRod1999: It looks like moving this to a helper function caused an issue where it seems like it is skipping it. The run time is only like 0.03 seconds. Unresolving comment for now

DanRod1999: fixed

packages/performance-tests/test/iModelUtils.ts
DanRod1999: done

packages/performance-tests/test/TransformerRegression.test.ts
DanRod1999: still had trouble with this, so may need to discuss more. However used this to start modifying reporter system

DanRod1999: done, syntax slightly different, so please check if that is what you intended

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -95,16 +100,69 @@ const setupTestData = async () => {
...
+            "Branch Name": branchName,
+            "Federation Guid Saturation 0-1": fedGuidSaturation,
+          };
+          sourceDb.close();
+        });
+
+        beforeEach(async () => {
+          assert(iModel._cachedFileName, "Imodel file name not Cached to TestIModel Object");
+          sourceDb = await BriefcaseDb.open({
+            fileName: iModel._cachedFileName,
```
MichaelBelousov, line 151: `_cachedFileName` should be private. Instead add a `let sourceFileName` variable and when you call `await iModel.getFileName()` in the `before` callback, instead do `sourceFileName = iModel.getFileName()` and just use that variable.

packages/performance-tests/test/TransformerRegression.test.ts
MichaelBelousov: reasonable.

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -95,16 +100,69 @@ const setupTestData = async () => {
...
+
+        afterEach(async () => {
+          sourceDb.close(); // closing to ensure connection cache reusage doesn't affect results
+        });
+
         testCasesMap.forEach(async (testCase, key) => {
           it(key, async () => {
-            reporter = await testCase(iModel, reporter, branchName);
+            await testCase(sourceDb,
+              (smallReportSubset: [string, string, string, number]) => {
```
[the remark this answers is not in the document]
MichaelBelousov: also double check we remove any options for the test case to pass data to the reporter that should not vary between multiple reports from the same test case

packages/performance-tests/test/TestContext.ts
```diff
@@ -49,16 +51,23 @@ export async function *getTestIModels(filter: (iModel: TestIModel) => boolean) {
         iModelId,
         iTwinId,
         tShirtSize: getTShirtSizeFromName(iModel.displayName),
-        load: async () => downloadAndOpenBriefcase({ iModelId, iTwinId }),
+        // _cachedFileName: undefined,
+        async getFileName(): Promise<void> {
```
MichaelBelousov, line 55: this should return the file name

packages/performance-tests/test/TestContext.ts
```diff
@@ -49,16 +51,23 @@ export async function *getTestIModels(filter: (iModel: TestIModel) => boolean) {
...
+          const _briefcase = await downloadBriefcase({ iModelId, iTwinId}); // note not downloadAndOpen
+          this._cachedFileName = _briefcase.fileName;
+        },
       };
       if(filter(iModelToCheck)){
```
MichaelBelousov, line 60:
```suggestion
      if (filter(iModelToCheck)) {
```

MichaelBelousov (approved):
lgtm, minor comments

packages/performance-tests/test/TransformerRegression.test.ts

packages/performance-tests/test/TransformerRegression.test.ts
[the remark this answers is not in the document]

packages/performance-tests/test/TestContext.ts
DanRod1999: done

packages/performance-tests/test/TestContext.ts
DanRod1999: done

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -95,16 +100,69 @@ const setupTestData = async () => {
...
+            "T-shirt size": iModel.tShirtSize,
+            "Gb size": sizeInGb,
+            "Branch Name": branchName,
+            "Federation Guid Saturation 0-1": fedGuidSaturation,
+          };
+          sourceDb.close();
+        });
+
+        beforeEach(async () => {
+          sourceFileName = await iModel.getFileName();
```
MichaelBelousov, line 149: do you really need to refresh the name? It's cached, and you already got it in your `before[all]` callback

MichaelBelousov (dismissed):

packages/performance-tests/test/TransformerRegression.test.ts
DanRod1999: good point done

MichaelBelousov (dismissed):

MichaelBelousov (approved):

Issue #83, opened 2023-06-26
Remove unbounded cache sizes


MichaelBelousov:
Unbounded caches in use are problematic:
- js Map in v8 (node) has a limit of 16.7M elements which has already been surpassed by Aspect counts in iModels
- large memory usage and garbage collector performance impacted by too many objects

Proposed changes:
- element remap table should be a temporary sqlite db, not a C++ data structure, so it can page to disk under memory pressure, and still be used in native code where geometry part remapping is done.
- aspect and element remap tables are probably, as the transformer uses them, good candidates for a custom data structure that stores "runs" (e.g. run-length encoding) of element mappings, (e.g. `7-10+10,15->16 === 7->17,8->18,9->19->10->20,15->16`).
- the entire pending reference cache can be removed by storing a smaller cache of just the ids of elements that had unresolved references, and then after exporting everything else which will have inserted all targeted elements, then do the second transform + update that currently occurs as soon as the references are found.

Short term solutions:
- Use BigMap classes for aspects.

MichaelBelousov:
worth noting, last time I tried using the integer based map in iTwin.js core for the transformer, it worsened performance 5-10% iirc. We need an integer based map though to do a naive run-compressed map. So need to think about this.

MichaelBelousov:
this is done partially in the `early-polymorphic-insert` branch

MichaelBelousov:
the federation-guid-optimization branch also adds a new source changes cache which is unbounded

PR #85, opened 2023-06-27
Start using BigMap instead of regular Map

simsta6:
When transforming iModels with huge amounts of Element Aspects process can crash.

BigMap implementation will decrease performance by little bit.

To decrease memory usage, we can use https://www.npmjs.com/package/lmdb

### DIFF
--- a/packages/transformer/src/BigMap.ts
+++ b/packages/transformer/src/BigMap.ts
@@ -9,7 +9,7 @@ import { Id64String } from "@itwin/core-bentley";
  * This class serves as a temporary solution to address the issue of the V8 JavaScript engine's Map implementation,
  * which is limited to approximately 16.7 million elements. Currently, only valid or invalid Id64 string
  * representations are supported by this Map; any other type will lead to errors. It's worth noting that
- * our future plan is to replace this stopgap with a more robust solution, utilizing a temporary SQLite database.
+ * our future plan is to replace this stopgap with a more robust solution, utilizing a temporary SQLite database (https://github.com/iTwin/imodel-transformer/issues/83).
  */
 export class BigMap<V> extends Map<Id64String, V> {
     private _maps: Record<string, Map<string, V>>;
--- a/packages/transformer/src/BigMap.ts
+++ b/packages/transformer/src/BigMap.ts
@@ -1,0 +1,74 @@
+const MAX_MAP_SIZE = 10_000_000;
+
+export class BigMap<K, V> extends Map<K, V> {
+  private _maps: Map<K, V>[];
+  private _size: number;
+
+  public override get size(): number {
+    return this._size;
+  }
+
+  public constructor() {
+    super();
+    this._maps = [new Map()];
+    this._size = 0;
+  }
+
+  public override clear(): void {
+    this._maps.forEach((m) => m.clear());
+    this._size = 0;
+  }
+
+  public override delete(key: K): boolean {
+    const wasDeleted = this._maps.some((m) => m.delete(key));
+    if (wasDeleted) {
+      this._size--;
+    }
+
+    return wasDeleted;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public override forEach(callbackfn: (value: V, key: K, map: Map<K, V>) => void, thisArg?: any): void {
+    this._maps.forEach((m) => {
+      m.forEach(callbackfn, thisArg);
+    });
+  }
+
+  public override get(key: K): V | undefined {
+    for (const map of this._maps) {
+      const value = map.get(key);
+      if (value !== undefined) {
+        return value;
+      }
+    }
+
+    return;
+  }
+
+  public override has(key: K): boolean {
+    return this._maps.some((m) => m.has(key));
+  }
+
+  public override set(key: K, value: V): this {
+    // duplicate key
+    for (const map of this._maps) {
+      if (map.has(key)) {
+        map.set(key, value);
+      }
+    }
+
+    const lastMap = this._maps[this._maps.length - 1];
+    if (lastMap.size < MAX_MAP_SIZE) { // last map has free space
+      lastMap.set(key, value);
+    } else { // item will be put into new map
+      const newMap = new Map();
+      newMap.set(key, value);
+      this._maps.push(newMap);
+    }
+
+    this._size++;
+
+    return this;
+  }
+}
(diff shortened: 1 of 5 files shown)


packages/transformer/src/BigMap.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    return;
+  }
+
+  public override has(key: K): boolean {
+    return this._maps.some((m) => m.has(key));
+  }
+
+  public override set(key: K, value: V): this {
```
MichaelBelousov, line 53: Would it be faster (and support approximately 167M elements) to have a submap bucket per last digit/character of the id64 during? Doesn't support as large of a list but will have O(1) access of buckets

MichaelBelousov:
> BigMap implementation will decrease performance by little bit.

Have you done any testing to quantify the difference? @DanRod1999 can you run the performance tests on this branch?

MichaelBelousov:
[internal performance report](https://app.powerbi.com/groups/2fc3ba7e-a05a-4cc0-9411-e9d2d2652ec1/reports/e684773d-e7aa-45ac-8e84-b33352436116/ReportSection?experience=power-bitests) says 2-3% increase in transformation time. Let's try the per-last-char-hashed bucket imo, but otherwise this is fine for a short term fix. Longer term fix will be trying to use a compressed map.

packages/transformer/src/BigMap.ts
KyryloVolotovskyi: It will only be faster if we are going to have 20+ million of aspects in one iModel. If there are <10 million aspects, the set method will only call `has` once. By checking the last digit/character of the id64, it will be more of an overhead than adding another `has` invocation.

The optimization that you have suggested, will only make sense for extremely large datasets and I think it would have a performance hit in common scenarios.

Do you think it is still worth to change it?

MichaelBelousov: I think it will still be faster because it's less complicated than the iterator logic here even when the iterator contains only one element. I think we should try it to see if it's better than the 2-3% performance hit in all transforms. Another option is make this class use a regular map until the count hits 16.7M elements, then add another and start using iterations.

MichaelBelousov:
> How many entities did the iModels that were used for testing have? If the amount was rather small then there shouldn't be a noticeable impact on the overall performance. Can 2-3% increase just be some noise?

@DanRod1999 can you answer how many entities are in the test iModels? Multiple iModels saw similar performance downgrades in the tests, and they are not small iModels. I do not think it was a fluke. Granted, my sample size is small, but when the tests run as long we have them, on those well-calibrated performance test machines, I think it's unlikely that was noise

DanRod1999:
> > How many entities did the iModels that were used for testing have? If the amount was rather small then there shouldn't be a noticeable impact on the overall performance. Can 2-3% increase just be some noise?
> 
> @DanRod1999 can you answer how many entities are in the test iModels? Multiple iModels saw similar performance downgrades in the tests, and they are not small iModels. I do not think it was a fluke. Granted, my sample size is small, but when the tests run as long we have them, on those well-calibrated performance test machines, I think it's unlikely that was noise

I don't know for sure but I can say the GB size for them are both less than 2GB. I believe the imodels I have been using are 1.2 and .87 gb. I will try to find the entity counts

packages/transformer/src/BigMap.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public override has(key: Id64String): boolean {
+      return this._maps[key[key.length - 1]].has(key);
+    }
+
+    public override set(key: Id64String, value: V): this {
+      const mapForKey = this._maps[key[key.length - 1]];
+      if (mapForKey === undefined)
+        throw Error(`Tried to set ${key} with ${value}`);
```
MichaelBelousov, line 70:
```suggestion
        throw Error(`Tried to set ${key}, but that key has no submap`);
```

packages/transformer/src/BigMap.ts
```diff
@@ -0,0 +1,77 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+import { Id64String } from "@itwin/core-bentley";
+
+export class BigMap<V> extends Map<Id64String, V> {
```
MichaelBelousov, line 8: document that this is a stopgap around the 16.7M limit of elements in the v8 implementation of JavaScript Map, and that we intend to replace the map usage with a temporary sqlite database anyway. You should link to #83

packages/transformer/src/BigMap.ts
KyryloVolotovskyi: Fixed

packages/transformer/src/BigMap.ts
KyryloVolotovskyi: Added docstring

packages/transformer/src/BigMap.ts
```diff
@@ -0,0 +1,83 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+import { Id64String } from "@itwin/core-bentley";
+
+/**
+ * This class serves as a temporary solution to address the issue of the V8 JavaScript engine's Map implementation,
+ * which is limited to approximately 16.7 million elements. Currently, only valid or invalid Id64 string
```
MichaelBelousov, line 10: "valid or invalid" is confusing. How about:
```suggestion
 * which is limited to approximately 16.7 million elements. Currently, only Id64 string
```

MichaelBelousov (approved):

packages/transformer/src/BigMap.ts
```diff
@@ -0,0 +1,83 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+import { Id64String } from "@itwin/core-bentley";
```
MichaelBelousov, line 6: you have a docs error because you don't have a docs tag for this module:
```
Topic BigMap does not have a @docs-group-description in imodel-transformer.ts. Please add a description.
```
Can we fix that?

MichaelBelousov (approved):

MichaelBelousov (approved):

PR #87, opened 2023-06-27
Implemented check for aspect identifier being not an id

KyryloVolotovskyi:


packages/transformer/src/IModelTransformer.ts
```diff
@@ -1107,6 +1107,9 @@ export class IModelTransformer extends IModelExportHandler {
       while (DbResult.BE_SQLITE_ROW === stmt.step()) {
         // ExternalSourceAspect.Identifier is of type string
         const aspectIdentifier = stmt.getValue(0).getString();
+        if(!Id64.isId64(aspectIdentifier)){
```
MichaelBelousov, line 1110: Do you have more information about the case in which you are seeing this? Technically, external source aspects with a scope of the targetScopeElement should have been placed by the transformer, and should have an identifier that is a valid hexadecimal number string for an iModel id, which is why previously it was assumed that the expensive isId64 check was not necessary.

MichaelBelousov: If you encountered an iModel where this is the case, then someone screwed up somewhere, but then we do need to handle it.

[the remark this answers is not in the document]
KyryloVolotovskyi: Ye, makes sense. Fixed it.

KyryloVolotovskyi: I am not aware of all the details but from what I understood, there are different iModels which are using `Identifier` as a reference to another iModel or file or whatever. Those 'references' are not any ids but some sort of paths to .dgn files. Here is the [query](https://qa-imodelconsole.bentley.com/?projectid=5ba46da3-9bfb-494a-beba-ae8f83309024&imodelid=64d75097-e515-4a93-a7e0-1b99da17222a&changesetid=71003a3a5b75c52bf5c31543ed6747b52072fba6&=SELECT%20*%20FROM%20BisCore.ExternalSourceAspect%20WHERE%20Scope.id=0x1%20AND%20Identifier%20like%20'%dgn%') which selects all of the `ExternalSourceAspect`'s in the iModel which will cause the `BE_SQLITE_ERROR: Error binding Id`.

PR #88, opened 2023-06-27
Fix element provenance direction in reverse synchronization workflow

ViliusRuskys:
`forEachTrackedElement` method iterates through related (tracked) elements in 2 imodels and invokes a callback on each pair.
There are 2 scenarios for iterating through tracked elements during transformer initialization.

1. Provenance is stored in target: Source ----> Target (prov)
2. Provenance is stored in source: Source (prov) ----> Target

These 2 scenarios only matter when the provenance is tracked by `ExternalSourceAspects` since depending on the provenance location the `element.id` and `identifier` fields will be switched.

When provenance is tracked by federation guids we don't care where the provenance is, because the callback should always be invoked in the data flow direction and not the provenance direction.
Methods that depend on the remap table (e.g. `findTargetElementId`) will only ever by called with the element id in the sourceDb to get element id in targetDb.

### DIFF
--- a/packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
+++ b/packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
@@ -1188,54 +1188,54 @@ describe("IModelTransformerHub", () => {
     assert.isTrue(Guid.isGuid(targetIModelId));
 
     try {
-        const sourceDb = await HubWrappers.downloadAndOpenBriefcase({ accessToken, iTwinId, iModelId: sourceIModelId });
-        const targetDb = await HubWrappers.downloadAndOpenBriefcase({ accessToken, iTwinId, iModelId: targetIModelId });
+      const sourceDb = await HubWrappers.downloadAndOpenBriefcase({ accessToken, iTwinId, iModelId: sourceIModelId });
+      const targetDb = await HubWrappers.downloadAndOpenBriefcase({ accessToken, iTwinId, iModelId: targetIModelId });
 
-        const categoryId = SpatialCategory.insert(sourceDb, IModel.dictionaryId, "C1", {});
-        const modelId = PhysicalModel.insert(sourceDb, IModel.rootSubjectId, "PM1");
-        const physicalElement: PhysicalElementProps = {
-            classFullName: PhysicalObject.classFullName,
-            model: modelId,
-            category: categoryId,
-            code: Code.createEmpty(),
-            userLabel: "Element1"
-        };
-        const originalElementId = sourceDb.elements.insertElement(physicalElement);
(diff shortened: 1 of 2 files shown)


packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -1179,6 +1179,66 @@ describe("IModelTransformerHub", () => {
...
+        sourceDb.saveChanges();
+        await sourceDb.pushChanges({ description: "change processing transformation" });
+
+        const masterElement = sourceDb.elements.getElement(originalElementId);
+        expect(masterElement).to.not.be.undefined;
+        expect(masterElement.userLabel).to.be.equal("Element1_updated");
+    } finally {
+        try {
+        // delete iModel briefcases
+        await IModelHost.hubAccess.deleteIModel({ iTwinId, iModelId: sourceIModelId });
```
MichaelBelousov, line 1233: indent this block two spaces... I really need to add prettier to the CI :(

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -1179,6 +1179,66 @@ describe("IModelTransformerHub", () => {
...
+        await IModelHost.hubAccess.deleteIModel({ iTwinId, iModelId: targetIModelId });
+        } catch (err) {
+        // eslint-disable-next-line no-console
```
MichaelBelousov, line 1236: indent here too

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -1179,6 +1179,66 @@ describe("IModelTransformerHub", () => {
...
+        await targetDb.pushChanges({ description: "initial transformation" });
+
+        const forkedElement = targetDb.elements.getElement(forkedElementId);
+        forkedElement.userLabel = "Element1_updated";
+        forkedElement.update();
+        targetDb.saveChanges();
+        await targetDb.pushChanges({ description: "update forked element's userLabel" });
+
+        transformer = new IModelTransformer(targetDb, sourceDb, {isReverseSynchronization: true});
+        await transformer.processChanges({startChangeset: {id: targetDb.changeset.id}});
```
MichaelBelousov, line 1223: nit: this should work
```suggestion
        await transformer.processChanges({startChangeset: targetDb.changeset});
```

MichaelBelousov (approved):

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
ViliusRuskys: sorry. I'm in the tab=4spaces gang so overlooked it :D. Fixed it now

PR #93, opened 2023-07-14
[fedGuid] Handle cases when element was recreated instead of updated

ViliusRuskys:
When an element gets updated by recreating it, the transformer sees 2 changes: 1 delete and 1 insert.
Since provenance is tracked by fed guids the insert will not happen, and an element update will be triggered instead. When this happens we no longer need to invoke the delete operation on the updated element.
We achieve this by keeping track of newly inserted elements in the target iModel and ignoring deletes on these elements.

### DIFF
--- a/packages/transformer/src/IModelExporter.ts
+++ b/packages/transformer/src/IModelExporter.ts
@@ -104,11 +104,6 @@ export abstract class IModelExportHandler {
    */
   public async preExportElement(_element: Element): Promise<void> {}
 
-  /** Look up a target ElementId from the source ElementId.
-   * @returns the target ElementId or [Id64.invalid]($bentley) if a mapping not found.
-   */
-  public findTargetElementId(_sourceElementId: Id64String): Id64String { return Id64.invalid; }
-
   /** Called when an element should be deleted. */
   public onDeleteElement(_elementId: Id64String): void { }
 
@@ -244,6 +239,26 @@ export class IModelExporter {
     this.sourceDb = sourceDb;
   }
 
+  /** state to prevent reinitialization of change data, @see [[initialize]] */
+  private _changeDataInitialized = false;
+
+  /**
+   * Initialize prerequisites of processing.
+   * If you wish to call [[exportChanges]] you must use identical [[ExportChangesOptions]] to the [[ExporterInitOptions.exportChangesOptions]] that you provided to [[initialize]] method.
+   * [[exportChanges]] will initialize the Exporter itself if it was not initialized prior to calling it so prefer using [[exportChanges]].
+   */
+  public async initialize(options: ExporterInitOptions): Promise<void> {
+    if (!this._changeDataInitialized && options?.exportChangesOptions && this.sourceDb.isBriefcaseDb()) {
+      await this.initializeChangeData({iModel: this.sourceDb, ...options.exportChangesOptions});
+      this._changeDataInitialized = true;
+    }
+  }
+
+  /** Initialize changeData that is needed to exportChanges */
+  private async initializeChangeData(args: ChangedInstanceIdsInitOptions): Promise<void> {
+    this._sourceDbChanges = await ChangedInstanceIds.initialize(args);
+  }
+
   /** Register the handler that will be called by IModelExporter. */
   public registerHandler(handler: IModelExportHandler): void {
     this._handler = handler;
@@ -284,6 +299,8 @@ export class IModelExporter {
    * @note [[exportSchemas]] must be called separately.
    */
   public async exportAll(): Promise<void> {
+    await this.initialize({});
+
     await this.exportCodeSpecs();
     await this.exportFonts();
     await this.exportModel(IModel.repositoryModelId);
@@ -307,20 +324,19 @@ export class IModelExporter {
       return;
     }
 
-    const opts: ExportChangesOptions
+    const initOpts: ExporterInitOptions
       = typeof accessTokenOrOpts === "object"
-        ? accessTokenOrOpts
+        ? { exportChangesOptions: accessTokenOrOpts }
         : {
-          accessToken: accessTokenOrOpts,
-          startChangeset: { id: startChangesetId },
+            exportChangesOptions: {
+              accessToken: accessTokenOrOpts,
+              startChangeset: { id: startChangesetId },
+            },
         };
 
-    this._sourceDbChanges =
-      (typeof accessTokenOrOpts === "object"
-        ? accessTokenOrOpts.changedInstanceIds
-        : undefined)
-      ?? await ChangedInstanceIds.initialize({ ...opts, iModel: this.sourceDb });
-    this.handleEntityRecreations();
+    await this.initialize(initOpts);
+    // _sourceDbChanges are initialized in this.initialize
+    nodeAssert(this._sourceDbChanges, "sourceDbChanges must be initialized.");
 
     await this.exportCodeSpecs();
     await this.exportFonts();
@@ -356,6 +372,9 @@ export class IModelExporter {
         this.handler.onDeleteRelationship(relInstanceId);
       }
     }
+
+    // Enable consecutive exportChanges runs without the need to re-instantiate the exporter.
+    this._changeDataInitialized = false;
   }
 
   /** Export schemas from the source iModel.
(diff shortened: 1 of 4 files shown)


packages/transformer/src/IModelTransformer.ts
```diff
@@ -307,6 +307,11 @@ export class IModelTransformer extends IModelExportHandler {
   /** a set of elements for which source provenance will be explicitly tracked by ExternalSourceAspects */
   protected _elementsWithExplicitlyTrackedProvenance = new Set<Id64String>();
 
+  /** a set of target element ids that were newly inserted during change processing workflow.
+   * This set will be initialized only after processing element inserts.
+   */
+  private _insertedTargetElementIds: Id64Set = new Set<Id64String> ();
```
MichaelBelousov, line 313:
```suggestion
  private _insertedTargetElementIds: Id64Set = new Set<Id64String>();
```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1464,12 +1469,21 @@ export class IModelTransformer extends IModelExportHandler {
     }
   }
 
+  /**
+   * Remember all target element ids that were newly inserted during current change processing.
+   */
+  public override async preDeleteElements(): Promise<void> {
+    this.exporter.sourceDbChanges?.element.insertIds.forEach((insertedSourceElementId) => this._insertedTargetElementIds.add(this.context.findTargetElementId(insertedSourceElementId)));
```
MichaelBelousov, line 1476: why introduce a new handler for this? Why not just a new private method in the `IModelTransformer` itself? I am hesitant to add new (even internal) handler methods, and I of the reasons I can think of that you have, I think there are better alternatives

MichaelBelousov (commented):
There is an unnecessary level of indirection here imo. While this does work, it leaves an unnecessary update in the cache, and also triggers unnecessary code (e.g. the delete that we then have to check for validity)

Imo a better solution would involve directly in the updated entities change data cache marking the element as not deleted, if another element was created with the same federation guid later. You could do that using the same technique you have here, where you generate a `Set` from the `sourceDbChanges.element.insertIds`, but instead of checking it during the deletion, check it during the updated elements cache creation so it is never considered a deletion

packages/transformer/src/IModelTransformer.ts
ViliusRuskys: needed that in order to access element remap table from exporter. Deleted this one and exposed `findTargetElementId` instead.

packages/transformer/src/IModelExporter.ts
```diff
@@ -104,6 +104,11 @@ export abstract class IModelExportHandler {
    */
   public async preExportElement(_element: Element): Promise<void> {}
 
+  /** Look up a target ElementId from the source ElementId.
+   * @returns the target ElementId or [Id64.invalid]($bentley) if a mapping not found.
+   */
+  public findTargetElementId(_sourceElementId: Id64String): Id64String { return Id64.invalid; }
```
MichaelBelousov, line 110: not sure the `IModelExportHandler` should really be aware of this... the `IModelTransformer` should be...

packages/transformer/src/IModelExporter.ts
```diff
@@ -315,6 +320,7 @@ export class IModelExporter {
         ? accessTokenOrOpts.changedInstanceIds
         : undefined)
       ?? await ChangedInstanceIds.initialize({ ...opts, iModel: this.sourceDb });
+    this.handleEntityRecreations();
```
MichaelBelousov, line 323: we need to think about how to do this in the transformer, not here. The exporter doesn't know/care about element "recreations" because it has no relationship to a target iModel, the transformer superimposes that relationship

MichaelBelousov (commented):
I think you moved it too far back :smile:

We don't want to bend the API so that the Exporter/ExportHandler is aware of transformations...

I meant let's do it in `IModelTransformer.remapDeletedSourceEntities` where we have access to change data, we can access the sourceDbChanges there.

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: resolving this as outdated, my new review addresses any concerns I have

ViliusRuskys:
The problem with doing it in the `IModelTransformer.remapDeletedSourceEntities` method is that the `exporter.sourceDbChanges` member is not yet initialized at that point. It is only 'for sure' initialized only after we call `exporter.exportChanges`.
I'm thinking of moving the sourceDbChanges initialization to a public `exporter.initializeChangeData()`. That way we could initialize the exporter  before running the `remapDeletedSourceEntities`.
What do you think about this approach?

MichaelBelousov:
> 

NOTE: discussed in private messages, I like this approach but by adding `IModelExporter.initialize` invoking `initializeChangeData` which is private

MichaelBelousov:
There is an optimization I had to remove to merge with main, I can reintroduce that optimization after your PR, since your PR (I believe) will move the `_sourceDbChanges` initialization to before `exportChanges`, and I need it in `IModelTransformer.processChanges` before `IModelExporter.exportChanges`, and I need access to it that early to reintroduce the optimization.
(which is to remap all non-deleted provenance-tracked elements ahead of time so that we can go back to the optimization of not exporting unchanged element in `processChanges`)

MichaelBelousov:
Also note @ViliusRuskys I see a failing test on the synced copy of this branch (idk if you have local work):

```mocha

  64 passing (1m)
  7 pending
  1 failing

  1) IModelCloneContext
       findTargetEntityId
         should return target relationship id:

      AssertionError: expected false to be true
      + expected - actual

      -false
      +true

      at Context.<anonymous> (src/test/standalone/IModelCloneContext.test.ts:124:47)
```

packages/transformer/src/IModelExporter.ts
```diff
@@ -302,19 +324,19 @@ export class IModelExporter {
       return;
     }
 
-    const opts: ExportChangesOptions
+    const initOpts: ExporterInitOptions
       = typeof accessTokenOrOpts === "object"
-        ? accessTokenOrOpts
+        ? { exportChangesOptions: accessTokenOrOpts }
```
MichaelBelousov, line 329: why add this wrapping subobject? Can't we just keep the properties inline (using `extends`)?

packages/transformer/src/IModelExporter.ts
```diff
@@ -870,6 +895,13 @@ export interface ChangedInstanceIdsInitOptions extends ExportChangesOptions {
   iModel: BriefcaseDb;
 }
 
+export interface ExporterInitOptions {
```
MichaelBelousov, line 898: if you add a type, mark it @beta for now.
Also this type has nothing in it... I agree we should add a new name, but make the properties inline:

```ts
export type ExporterInitOptions = ExportChangesOptions;
```
Later if we add more properties we can do:
```ts
export interface ExporterInitOptions extends ExportChangesOptions {
  newProperty: string;
}
```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -879,6 +892,14 @@ export class IModelTransformer extends IModelExportHandler {
               continue;
 
             this.context.remapElement(instId, targetId);
+            // If an entity insert and an entity delete both point to the same entity in target iModel, that means that entity was recreated.
+            // In such case an entity update will be triggered and we no longer need to delete the entity.
+            if(alreadyImportedElementInserts.has(targetId)) {
```
MichaelBelousov, line 897:
```suggestion
            if (alreadyImportedElementInserts.has(targetId)) {
```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -879,6 +892,14 @@ export class IModelTransformer extends IModelExportHandler {
...
+              this.exporter.sourceDbChanges?.element.deleteIds.delete(instId);
+            }
+            if(alreadyImportedModelInserts.has(targetId)) {
```
MichaelBelousov, line 900:
```suggestion
            if (alreadyImportedModelInserts.has(targetId)) {
```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1980,6 +2001,12 @@ export class IModelTransformer extends IModelExportHandler {
 
     await this.context.initialize();
     await this._tryInitChangesetData(args);
+
+    const exporterInitOptions: ExporterInitOptions = {};
+    if(this._isSynchronization)
```
MichaelBelousov, line 2006:
```suggestion
    if (this._isSynchronization)
```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2375,6 +2395,17 @@ export class IModelTransformer extends IModelExportHandler {
     this.finalizeTransformation();
   }
 
+  /** Changeset data must be initialized in order to build correct changeOptions. Call [[IModelTransformer.initialize]] for initialization of synchronization provenance data */
+  private getExportChangesOptions(accessToken?: AccessToken, startChangesetId?: string): ExportChangesOptions {
```
MichaelBelousov, line 2399: nit: worth thinking about if we can assert something to be sure this is called after change data is processed... like add a new private variable that is set to true after changedata is processed... or make this._changesetRanges have a third empty state

packages/transformer/src/IModelExporter.ts
ViliusRuskys: For future proofing.
If there will be a need to have init options for `exportAll` workflow in the future, then we can have separate members for both (`exportAll` and `exportChanges`) workflows. Since these are different workflows, we won't ever need to initialize both of them at the same time.
If we go with the `extends` now, then all the options would be in the same "bucket" and it would be hard to decide which options you actually need when you wanted to call `initialize`.

MichaelBelousov:
Since you are out @ViliusRuskys I will make the changes I requested in a separate branch and merge them. Please feel free to debate me on some ofmy changes and we can consider reverting them.

MichaelBelousov:
succeeded by https://github.com/iTwin/imodel-transformer/pull/99

PR #100, opened 2023-07-27
Added isFirstSynchronization to options

simsta6:
- Added isFirstSynchronizations to options
- Made shouldDetectDeletes method protected

If source iModel had different external source aspects, they were imported to target. Then at the end of transformation `detectElementsDeletes` was called. It would detect that some elements should be deleted, because transformer would get random identifier values, example:
image

### DIFF
--- a/change/@itwin-imodel-transformer-5c762aed-a0d1-466c-9388-7d9a060b3c4a.json
+++ b/change/@itwin-imodel-transformer-5c762aed-a0d1-466c-9388-7d9a060b3c4a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Added 'isFirstSynchronization' to options",
+  "packageName": "@itwin/imodel-transformer",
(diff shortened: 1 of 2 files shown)


MichaelBelousov:
> If source iModel had different external source aspects, they were imported to target. Then at the end of transformation `detectElementsDeletes` was called. It would detect that some elements should be deleted, because transformer would get random identifier values, example: image

We shouldn't introduce this "new option" because the transformer is going to expose a higher-level Branching API soon, including an `initBranch` API which would supersede this. Also, it appears that you're adding this because you're not using `wasSourceIModelCopiedToTarget` in the branch initialization, which is not supported by the transformer.

MichaelBelousov:
Also adding, it seems the bug this was fixing was due to a default targetScopeElement `0x1` which conflicted with the scope of ExternalSourceAspect's added by others. The default of `targetScopeElement` is very bad and the default will be removed in `1.x`

PR #102, opened 2023-07-27
Added functionality to dynamically load and test extra transformers in performance tests

KyryloVolotovskyi:


packages/performance-tests/test/ProgressTransformer.ts
```diff
@@ -0,0 +1,43 @@
...
+      return super.onExportElement(sourceElement);
+    }
+    public override onExportRelationship(sourceRelationship: Relationship): void {
+      this._increment();
+      return super.onExportRelationship(sourceRelationship);
+    }
+  }
+
+
+const progressTransformerTestModule: TestTransformerModule = {
```
MichaelBelousov, line 32: I don't think the 'progress' part is important enough to include in the name. It's really the 'native' or 'naked' transformer. The progress is just a side effect we include for logging, unimportant to its role

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -38,6 +40,22 @@ const testCasesMap = new Map([
...
 const outputDir = path.join(__dirname, ".output");
 
+const loadTransformers = async () => {
+  const moduleNames = process.env.EXTRA_TRANSFORMERS?.split(',').map((name) => name.trim()) ?? [];
+  const transformerModules = [ProgressTransformer];
+
+  for (const moduleName of moduleNames) {
+    try {
+      const module = (await import(moduleName)).default;
+      transformerModules.push(module.default);
```
MichaelBelousov, line 50: do we really need two levels of default? I encountered this before... is it due to esmodule interop?

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -156,13 +175,16 @@ async function runRegressionTests() {
...
         });
 
-        testCasesMap.forEach(async (testCase, key) => {
-          it(key, async () => {
-            await testCase(sourceDb,
-              (...smallReportSubset: [testName: string, iModelName: string, valDescription: string, value: number]) => {
-                reporter.addEntry( ...smallReportSubset, reportInfo );
-              });
-          }).timeout(0);
+        transformerModules.forEach((transformerModule: TestTransformerModule) => {
```
MichaelBelousov, line 178: let's flip the order of these two for loops, ideally we run every transformer Module for each test case, then move onto the next test case

packages/performance-tests/test/cases/prepare-fork.ts
```diff
@@ -116,13 +117,7 @@ async function classicalTransformerBranchInit(sourceDb: BriefcaseDb, branchDb: S
...
 
   // initialize the branch provenance
-  const branchInitializer = new IModelTransformer(sourceDb, branchDb, {
-    // tells the transformer that we have a raw copy of a source and the target should receive
-    // provenance from the source that is necessary for performing synchronizations in the future
-    wasSourceIModelCopiedToTarget: true,
-    // store the synchronization provenance in the scope of our representation of the external source, master
-    targetScopeElementId: masterExternalSourceId,
-  });
+  const branchInitializer = await transformerModule.createForkInitTransform!(sourceDb, branchDb);
```
MichaelBelousov, line 120: probably better assert with a verbose error that it exists

packages/performance-tests/test/TestTransformerNodule.ts
```diff
@@ -0,0 +1,11 @@
+import { IModelDb } from "@itwin/core-backend";
+import { IModelTransformer } from "@itwin/imodel-transformer";
+
+/**
+ * the type of a module imported by the tests holding a custom transformer
+ * implementation to test
+ */
+export interface TestTransformerModule {
+  createIdentityTransform?(sourceDb: IModelDb, targetDb: IModelDb): Promise<IModelTransformer>;
```
MichaelBelousov, line 9: didn't we want to change these from returning an IModelTransformer to a generic `{ run: () => Promise<void> }` object?

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -38,6 +40,22 @@ const testCasesMap = new Map([
...
 
+const loadTransformers = async () => {
+  const moduleNames = process.env.EXTRA_TRANSFORMERS?.split(',').map((name) => name.trim()) ?? [];
+  const transformerModules = new Map<string, TestTransformerModule>();
+  transformerModules.set("NativeTransformer", nativeTransformerModule);
+  try {
+    const modules = await Promise.all(
+      moduleNames.map(async moduleName => ({ name: moduleName, module: (await import(moduleName)).default }))
+    );
+    modules.forEach(({ name, module }) => transformerModules.set(name, module));
```
MichaelBelousov, line 51: I think this would be a bit cleaner without the extra steps and possibly in a more declarative approach:
```ts
// renamed to paths
const modulePaths = process.env.EXTRA_TRANSFORMERS?.split(',').map((name) => name.trim()) ?? [];

const envSpecifiedExtraTransformerCases = await Promise.all(
  modulePaths.map(async (m) => [m, (await import(m).catch(logError)).default])
);

const transformerModules = new Map<string, TestTransformerModule>([
  ["NativeTransformer", nativeTransformerModule],
  ...envSpecifiedExtraTransformerCases
]);
```

MichaelBelousov: also do we even need to catch module load error? They'll be rare and user configuration error that isn't super hard to figure out when presented with the crashing callstack.
I'd rather fail the test than keep going if we can't find a specified test transformer

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -157,12 +176,18 @@ async function runRegressionTests() {
...
-          it(key, async () => {
-            await testCase(sourceDb,
-              (...smallReportSubset: [testName: string, iModelName: string, valDescription: string, value: number]) => {
-                reporter.addEntry( ...smallReportSubset, reportInfo );
+          transformerModules.forEach((transformerModule: TestTransformerModule, moduleName: string) => {
+            it(`${key} on ${moduleName}`, async () => {
+              await testCase({
+                sourceDb,
+                transformerModule,
+                addReport: (...smallReportSubset: [testName: string, iModelName: string, valDescription: string, value: number]) => {
```
MichaelBelousov, line 184: if `testCase` is properly typed, you shouldn't need to explicitly specify the type of the rest argument

packages/performance-tests/test/cases/TestCaseContext.ts
```diff
@@ -0,0 +1,8 @@
+import { BriefcaseDb } from "@itwin/core-backend";
+import { TestTransformerModule } from "../TestTransformerNodule";
+
+export type TestCaseContext = {
+    sourceDb: BriefcaseDb;
+    transformerModule: TestTransformerModule;
+    addReport: (...smallReportSubset: [testName: string, iModelName: string, valDescription: string, value: number]) => void;
```
MichaelBelousov, line 7: I didn't like this original type much a while ago but now that we're moving things around again, extract this complicated type to like `ReportCallback` without the redundant spread-tuple:
```ts
type ReportCallback =  (testName: string, iModelName: string, valDescription: string, value: number) => void;
```

packages/performance-tests/test/TestTransformerNodule.ts
MichaelBelousov, line 1: this file should be named `Module` not `Nodule` :)

packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -9,35 +9,53 @@
...
+const loadTransformers = async () => {
+  const modulePaths = process.env.EXTRA_TRANSFORMERS?.split(",").map((name) => name.trim()).filter(Boolean) ?? [];
+  const envSpecifiedExtraTransformerCases = await Promise.all(
+    modulePaths.map(async (m) => [m, (await import(m)).default])
+  );
+  const transformerModules = new Map<string, TestTransformerModule>([
+    ["NativeTransformer", nativeTransformerTestModule],
+    ["RawForkOperations", rawForkOperationsTestModule],
+    ["RawForkCreateFedGuids", rawForkCreateFedGuidsTestModule],
+    ...envSpecifiedExtraTransformerCases as [string, TestTransformerModule][],
```
MichaelBelousov, line 54: move this cast to the declaration of `envSpecifiedExtraTransformerCases`

packages/performance-tests/test/cases/TestCaseContext.ts
```diff
@@ -0,0 +1,10 @@
+import { BriefcaseDb } from "@itwin/core-backend";
+import { TestTransformerModule } from "../TestTransformerNodule";
+
+type ReportCallback =  (testName: string, iModelName: string, valDescription: string, value: number) => void;
```
MichaelBelousov, line 4:
```suggestion
type ReportCallback = (testName: string, iModelName: string, valDescription: string, value: number) => void;
```

MichaelBelousov (commented):
Looks pretty good so far! Minor comments. Also @DanRod1999 will need to review for not breaking the test pipeline and performance report.

DanRod1999:
> Looks pretty good so far! Minor comments. Also @DanRod1999 will need to review for not breaking the test pipeline and performance report.

looks good to me. @KyryloVolotovskyi if you could send me the report.csv of a completed run so I can double check the format that would be great.

KyryloVolotovskyi:
> > Looks pretty good so far! Minor comments. Also @DanRod1999 will need to review for not breaking the test pipeline and performance report.
> 
> looks good to me. @KyryloVolotovskyi if you could send me the report.csv of a completed run so I can double check the format that would be great.

Sent in personal messages

MichaelBelousov (approved):
LGTM but let's not merge until you've (@DanRod1999?) kicked off a successful pipeline run

PR #104, opened 2023-08-02
Fixed onDeleteModel "Missing Id" and "ForeignKey constraint" errors

DeividasDavidav:
"Missing Id: error deleting model" error:
Added a fix and test for "Missing Id: error deleting model" which occured because onDeleteModel deleted definition model. It shouldn't delete definition model in onDeleteModel if its partition element will also be deleted, because elementTreeDeleter deletes definition model while deleting definition partition element.

"ForeignKey Constraint: error deleting model" error:
Added a fix and test for this error. Transformer tries to delete models before it deletes elements. Definition models cannot be deleted unless all of their modeled elements are deleted first.

### DIFF
--- a/packages/transformer/src/IModelTransformer.ts
+++ b/packages/transformer/src/IModelTransformer.ts
@@ -983,8 +983,37 @@ export class IModelTransformer extends IModelExportHandler {
     // - If only the model is deleted, [[initFromExternalSourceAspects]] will have already remapped the underlying element since it still exists.
     // - If both were deleted, [[remapDeletedSourceElements]] will find and remap the deleted element making this operation valid
     const targetModelId: Id64String = this.context.findTargetElementId(sourceModelId);
-    if (Id64.isValidId64(targetModelId)) {
-      this.importer.deleteModel(targetModelId);
+    try {
+      if (Id64.isValidId64(targetModelId)) {
+        if (this.exporter.sourceDbChanges?.element.deleteIds.has(sourceModelId)) {
+          const targetPartitionElement = this.targetDb.elements.tryGetElement(targetModelId);
+          const isDefinitionPartition = targetPartitionElement && targetPartitionElement instanceof DefinitionPartition;
+          if (isDefinitionPartition) {
+            // Skipping model deletion because model's partition will also be deleted.
+            // It expects that model will be present and will fail if it's missing.
+            // Model will be deleted when its partition will be deleted.
+            return;
+          }
+        }
+        this.importer.deleteModel(targetModelId);
+      }
+    } catch (error) {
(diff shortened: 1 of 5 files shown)


packages/transformer/src/IModelTransformer.ts
```diff
@@ -983,8 +983,37 @@ export class IModelTransformer extends IModelExportHandler {
...
+      // Transformer tries to delete models before it deletes elements. Definition models cannot be deleted unless all of their modeled elements are deleted first.
+      // In case a definition model needs to be deleted we need to skip it for now and register its modeled partition for deletion.
+      // The `OnDeleteElement` calls `DeleteElementTree` Which deletes the model together with its partition after deleting all of the modeled elements.
+      this.scheduleModeledPartitionDeletion(sourceModelId);
+    }
+  }
+
+  /** Schedule modeled partition deletion */
+  private scheduleModeledPartitionDeletion(sourceModelId: Id64String): void {
+    const deletedElements = this.exporter.sourceDbChanges?.element.deleteIds as Set<Id64String>;
```
MichaelBelousov, line 1014: this can error if there were no changed elements (you may want to add a test for that case). Don't cast away the undefined, instead return early if it is undefined or create the deleteIds set and add to it

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -857,5 +857,75 @@ describe("IModelTransformerHub", () => {
...
     sinon.restore();
   });
+
+  it("should delete definition elements and models when processing changes", async () => {
+    let definitionPartitionId1: string;
+    let definitionModelId1: string;
+    let definitionContainerId1: string;
+    let definitionModelId2: string;
+
+    const timeline: Timeline = {
```
MichaelBelousov, line 867: nit: can use an array here if you prefer. I only used an explicitly integer-indexed object in an older version of this test utility but it's still usable so not necessary to change

DeividasDavidav: Since it's still usable, I will leave it like it is

packages/transformer/src/IModelTransformer.ts
DeividasDavidav: If we only delete the model of a partition element, it shouldn't throw an error and "scheduleModeledPartitionDeletion" wouldn't be used. Also, it seems like if there are no changed elements, .element.deleteIds is defined, just empty.

MichaelBelousov (commented):
looks like I never submitted this review... sorry about that. Working on it now since your term ended

PR #108, opened 2023-08-07
also track source's sync changesets during reverse sync

MichaelBelousov:
also track source's sync changesets during reverse sync


MichaelBelousov:
should probably add a test where provenance during a reverse sync causes a changeset on the source.

ViliusRuskys (approved):

PR #111, opened 2023-08-16
Dan/fix-reporter

DanRod1999:
Add moduleName to TestSuite section of reporter


packages/performance-tests/test/TransformerRegression.test.ts
```diff
@@ -175,13 +175,13 @@ async function runRegressionTests() {
...
+        testCasesMap.forEach(async ({testCase, functionNameToValidate}, testCaseName) => {
           transformerModules.forEach((transformerModule: TestTransformerModule, moduleName: string) => {
             const moduleFunc = transformerModule[functionNameToValidate as keyof TestTransformerModule];
             if (moduleFunc) {
-              it(`${key} on ${moduleName}`, async () => {
-                const addReport = (testName: string, iModelName: string, valDescription: string, value: number) => {
-                  reporter.addEntry(testName, iModelName, valDescription, value, reportInfo);
+              it(`${testCaseName} on ${moduleName}`, async () => {
+                const addReport = (iModelName: string, valDescription: string, value: number) => {
+                  reporter.addEntry(`${testCaseName} ${moduleName}`, iModelName, valDescription, value, reportInfo);
```
MichaelBelousov, line 184: we are also iterating over `transformerModules`, we need that to be data that the test knows about too, otherwise how is the report supposed to know the difference between the native transformer running a test, and the service transformer, or the new branch init API? Idk how it's done today but it's weird to me that information isn't directly passed to the reporter as part of the entry.

MichaelBelousov: ok I understand now, my bad. I got confused by `moduleName`, I personally refer to it as like "a transformer subclass" or something, but our naming is inconsistent so it's hard.
Also I was confused by you removing the testName (which is the right thing to do) because I didn't notice that it was now redundant.

This is correct then, resolving.

MichaelBelousov (approved):

PR #113, opened 2023-08-21
Add detached ElementAspect processing

mindaugasdirg:
Changes:
- Add detached ElementAspect exporting.
- Extract ElementAspect exporting from IModelExporter.
- Add `elementAspectExportStrategy` option for IModelTransformer for choosing which ElementExporting strategy should be used during the transformation.
- Add BigMap usage to aspect remap table to remove ~17 million ElementAspect limit in iModels.
- Add ECSqlReaderAsyncIteratorAdapter for iTwinJs 3.x versions


packages/transformer/src/IModelExporter.ts
```diff
@@ -220,26 +239,54 @@ export class IModelExporter {
...
    * @see registerHandler
    */
-  public constructor(sourceDb: IModelDb) {
+  public constructor(sourceDb: IModelDb, elementAspectsStrategy: ElementAspectExportStrategy = ElementAspectExportStrategy.WithElement) {
     this.sourceDb = sourceDb;
+    this._exportElementAspectsStrategy = this.createExportElementAspectsStrategy(sourceDb, elementAspectsStrategy);
+  }
+
+  private createExportElementAspectsStrategy(sourceDb: IModelDb, elementAspectsStrategy: ElementAspectExportStrategy): ExportElementAspectsStrategy<ElementAspectsHandler> {
+    switch(elementAspectsStrategy) {
```
MichaelBelousov, line 258: instead of an enum, why not just take the actual class to call new on? (as an argument to this and the constructor)
```ts
// contrived, probably needs adjustments
interface ElementAspectExportStrategyHandler {
  trackProgress: () => Promise<void>, // ew :/ can't this be made sync?
  onExportElementMultiAspects: (multiAspects) => Id64String[] | void,
  onExportElementUniqueAspect: (uniqueAspect, isUpdate) => Id64String | void,
  shouldExportElementAspect: (aspect: ElementAspect) => boolean,
}
//...

private createExportElementAspectsStrategy(
  sourceDb: IModelDb,
  elementAspectsStrategy: new (
    src: IModelDb,
    handler: ElementAspectExportStrategyHandler
  ) => ExportElementAspectsStrategy
) {
  return new HandlerClass(src, {
    trackProgress: async () => this.trackProgress(),
    onExportElementMultiAspects: (multiAspects) => this.handler.onExportElementMultiAspects(multiAspects),
    onExportElementUniqueAspect: (uniqueAspect, isUpdate) => this.handler.onExportElementUniqueAspect(uniqueAspect, isUpdate),
    shouldExportElementAspect: (aspect) => this.handler.shouldExportElementAspect(aspect),
  });
}

```

packages/transformer/src/IModelExporter.ts
```diff
@@ -220,26 +239,54 @@ export class IModelExporter {
...
 
   /** Register the handler that will be called by IModelExporter. */
   public registerHandler(handler: IModelExportHandler): void {
     this._handler = handler;
   }
 
+  /** Sets ExportElementAspectsStrategy that will be used when exporting ElementAspects.
+   * This method allows to change strategy after IModelExporter was already created.
+   */
+  public setExportElementAspectsStrategy(elementAspectsStrategy: ElementAspectExportStrategy): void {
```
MichaelBelousov, line 286: is this new API necessary? What happens if you switch from "export per element" to "export all" in the middle? You re-export conflicting aspects?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -309,6 +316,7 @@ export class IModelTransformer extends IModelExportHandler {
     }
     this.sourceDb = this.exporter.sourceDb;
     this.exporter.registerHandler(this);
+    this.exporter.setExportElementAspectsStrategy(this._options.elementAspectExportStrategy ?? ElementAspectExportStrategy.WithElement);
```
MichaelBelousov, line 319: I would prefer we allow people passing in a custom exporter to set this themselves, and we can just log a warning (or throw an error) if they pass in a custom exporter at the same time as passing not "undefined" to the transformer options.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1240,6 +1248,10 @@ export class IModelTransformer extends IModelExportHandler {
     return targetRelationshipProps;
   }
 
+  public override shouldExportElementAspect(aspect: ElementAspect) {
+    return this.context.findTargetElementId(aspect.element.id) !== Id64.invalid;
```
MichaelBelousov, line 1252: This seems incorrect... `shouldExportElementAspect` should be based on the aspect itself, not the transformer's internal processing state of that aspect. Is it possible to remove this requirement?

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -2512,6 +2512,57 @@ describe("IModelTransformer", () => {
...
+    targetDb.saveChanges();
+
+    // assert
+    const elementIds = targetDb.queryEntityIds({ from: Subject.classFullName });
+    elementIds.forEach((elementId) => {
+      if (elementId === IModel.rootSubjectId) {
+        return;
+      }
+      const aspects = targetDb.elements.getAspects(elementId, ExternalSourceAspect.classFullName);
+      expect(aspects.length).to.be.equal(6); // +1 because provenance aspect was added
```
MichaelBelousov, line 2562: instead of hardcoding, can you check that the aspects.length is equal to the source element's aspect.length + 1?

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -797,6 +794,71 @@ describe("IModelTransformerHub", () => {
...
+            kind: "Document",
+            scope: { id: IModel.rootSubjectId, relClassName: "BisCore:ElementScopesExternalSourceIdentifier" },
+          };
+
+          const aspectId = sourceSeedDb.elements.insertAspect(aspectProps);
+          aspectIds.push(aspectId); // saving for later deletion
+        }
+      });
+
+      return Promise.resolve();
```
MichaelBelousov, line 822: why return `Promise.resolve()` explicitly in an async function? You don't even use `await`, so you could just make the caller allow non-promise results, and make this function not-async

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -797,6 +794,71 @@ describe("IModelTransformerHub", () => {
...
+      await saveAndPushChanges(sourceDb, "Update source");
+
+      await transformer.processChanges({ accessToken, startChangeset: sourceDb.changeset });
+      await saveAndPushChanges(targetDb, "Second transformation");
+
+      const targetElementIds = targetDb.queryEntityIds({ from: Subject.classFullName, where: "EcInstanceId != ?", bindings: [IModel.rootSubjectId] });
+      targetElementIds.forEach((elementId) => {
+        const aspects = targetDb.elements.getAspects(elementId, ExternalSourceAspect.classFullName) as ExternalSourceAspect[];
+        expect(aspects.length).to.be.equal(1); // +1 because provenance aspect was added
+        const aspectAddedAfterFirstTransformation = aspects.find((aspect) => aspect.identifier === "aspectAddedAfterFirstTransformation");
```
MichaelBelousov, line 853: O_o where does this come from?

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -797,6 +794,71 @@ describe("IModelTransformerHub", () => {
...
+
+          const aspectId = sourceSeedDb.elements.insertAspect(aspectProps);
+          aspectIds.push(aspectId); // saving for later deletion
+        }
+      });
+
+      return Promise.resolve();
+    });
+
+    const targetIModelId = await createPopulatedIModelHubIModel("TransformerTarget", async () => Promise.resolve());
```
MichaelBelousov, line 825: make the second argument optional

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -57,34 +57,31 @@ describe("IModelTransformerHub", () => {
   });
   after(() => HubMock.shutdown());
 
+  const createPopulatedIModelHubIModel = async (iModelName: string, prepareIModel: (iModel: SnapshotDb) => Promise<void>): Promise<string> => {
```
MichaelBelousov, line 60:
```suggestion
  const createPopulatedIModelHubIModel = async (
    iModelName: string,
    prepareIModel?: (iModel: SnapshotDb) => void | Promise<void>
  ): Promise<string> => {
```
make second argument optionally async and in general optional

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -57,34 +57,31 @@ describe("IModelTransformerHub", () => {
...
+    // Create and push seed of IModel
+    const seedFileName = path.join(outputDir, `${iModelName}.bim`);
+    if (IModelJsFs.existsSync(seedFileName))
+      IModelJsFs.removeSync(seedFileName);
+
+    const seedDb = SnapshotDb.createEmpty(seedFileName, { rootSubject: { name: iModelName } });
+    assert.isTrue(IModelJsFs.existsSync(seedFileName));
+    await prepareIModel(seedDb);
```
MichaelBelousov, line 68: execute optionally:
```suggestion
    await prepareIModel?.(seedDb);
```

MichaelBelousov:
Forgot to put as my review summary comment, there are some really great refactors in here! Thank you for working on this! Overall really good, but I have several nits and some small changes we can discuss.

packages/transformer/src/IModelCloneContext.ts
```diff
@@ -28,7 +29,7 @@ const loggerCategory: string = TransformerLoggerCategory.IModelCloneContext;
 export class IModelCloneContext extends IModelElementCloneContext {
 
   private _refTypesCache = new ECReferenceTypesCache();
-  private _aspectRemapTable = new Map<Id64String, Id64String>();
+  private _aspectRemapTable = new BigMap<Id64String>();
```
MichaelBelousov, line 32: Regarding this, while I would defer the work to another PR, but I did realize we probably can in javascript compress the remap table with runs pretty easily, by just inlining the two numbers directly into an array and running an adjusted binary search on that array.

mindaugasdirg: I think a better approach would be to minimize how many aspects are being saved in this table. Aspects cannot be referenced by other elements in EC schemas and as far as I know elements should reference owning element instead as aspect is sort of extension of element. So this table should only be needed when partially inserted aspect is fully completed so that transformer could update it. After that it should be safe to remove aspect's record from the table and in case aspect doesn't have pending references it should be able to entirely skip storing it in this remap table.

But I agree that this should be a separate PR and removed this change from here.

packages/transformer/src/ExportElementAspectsStrategy.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+    // ElementAspect has passed standard exclusion rules, now give handler a chance to accept/reject
+    return this.handler.shouldExportElementAspect(aspect);
+  }
+
+  public setAspectChanges(aspectChanges?: ChangedInstanceOps) {
+    this.aspectChanges = aspectChanges;
+  }
+
+  public getExcludedElementAspectClasses(): Set<string> {
```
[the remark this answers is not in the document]
mindaugasdirg: That would be better, but I need to keep support for loading excluded ElementAspect classes during transformation resumption. In that use case Transformer classes are already created when the state is loaded. I can convert this method to getter property instead if that works with you.

```ts
public get excludedElementAspectClasses(): Set<typeof ElementAspect> {
    this._excludedElementAspectClassFullNames;
}
```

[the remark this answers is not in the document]
MichaelBelousov: Resumption is deprecated, and we can just use `as any` in the resumption rehydration step imo. But feel free to do as you say and I'll re-review

packages/transformer/src/IModelCloneContext.ts
MichaelBelousov: yes you're right that should be possible, resolving

packages/transformer/src/DetachedExportElementAspectsStrategy.ts
```diff
@@ -0,0 +1,108 @@
...
+    for await (const rowProxy of aspectClassesAsyncQueryReader) {
+      const row = rowProxy.toRow();
+      aspectClassNameIdMap.set(row.className, row.classId);
+    }
+
+    for (const [className, classId] of aspectClassNameIdMap) {
+      if(this._excludedElementAspectClassFullNames.has(className))
+        continue;
+
+      const getAspectPropsSql = `SELECT * FROM ${className} WHERE ECClassId = :classId ORDER BY Element.Id`;
```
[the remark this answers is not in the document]
mindaugasdirg: It might be, but the reason for this ordering is that because `IModelImporter` and from it `IModelTransformer` requires that all element's `ElementMultiAspects` with the same class would be processed in one bunch. Ordering aspects by `ECClassId, ECInstanceId` won't ensure that all element's aspects of the same class will be exported in one bunch.

Ideally I would like to refactor that logic as well, but do it as a separate PR since there are more bugs that are present in that logic which are not related to memory optimizations or how exporter exports aspects (for example, it will miss to delete aspects in `targetDb`, if during `processChanges` there are no longer any aspects of that class in the `sourceDb`)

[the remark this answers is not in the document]
MichaelBelousov: reasonable, thanks, closing

packages/transformer/src/ExportElementAspectsStrategy.ts
```diff
@@ -0,0 +1,74 @@
...
+  protected aspectChanges: ChangedInstanceOps | undefined;
+
+  protected handler: T;
+
+  public constructor(sourceDb: IModelDb, handler: T) {
+    this.sourceDb = sourceDb;
+    this.handler = handler;
+  }
+
+  public abstract exportElementAspects(elementId: Id64String): Promise<void>;
```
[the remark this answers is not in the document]
mindaugasdirg: Renamed it, just did it in a separate commit so that it would also rename all references to the method

packages/transformer/src/DetachedExportElementAspectsStrategy.ts
```diff
@@ -0,0 +1,108 @@
...
+      const aspectQueryReader = this.sourceDb.createQueryReader(getAspectPropsSql, new QueryBinder().bindId("classId", classId), { rowFormat: QueryRowFormat.UseJsPropertyNames });
+      const aspectAsyncQueryReader = ensureECSqlReaderIsAsyncIterableIterator(aspectQueryReader);
+      let firstDone = false;
+      for await (const rowProxy of aspectAsyncQueryReader) {
+        const row = rowProxy.toRow();
+        const aspectProps: ElementAspectProps = { ...row, classFullName: row.className.replace(".", ":"), className: undefined }; // add in property required by EntityProps
```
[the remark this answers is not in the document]
mindaugasdirg: Thanks, I've used in the an earlier query which gathers aspect classes and using gathered aspect class name here.

packages/transformer/src/DetachedExportElementAspectsStrategy.ts
```diff
@@ -0,0 +1,108 @@
...
+      }
+
+      await exportAspect(aspect);
+    }
+  }
+
+  private async *queryAspects<T extends ElementAspect>(baseElementAspectClassFullName: string) {
+    const aspectClassNameIdMap = new Map<string, Id64String>();
+
+    const getAspectClassesSql = `SELECT DISTINCT ECClassId as classId, (ec_classname(ECClassId)) as className FROM ${baseElementAspectClassFullName}`;
```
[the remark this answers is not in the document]
mindaugasdirg: Good catch, I've rewrote the query to not have `SELECT DISTINCT` and used the same approach like `ECReferenceCache` to scan `ECDbMeta` schema to gather aspect classes.

packages/transformer/src/DetachedExportElementAspectsStrategy.ts
```diff
@@ -0,0 +1,108 @@
...
+    const aspectClassesQueryReader = this.sourceDb.createQueryReader(getAspectClassesSql);
```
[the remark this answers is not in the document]
mindaugasdirg: Yeah, I think we don't need to use primary connection here. Source shouldn't be changed here as it won't be updated since transformer only updates `targetDb` and even if `provenanceDb` would be `sourceDb`, provenance is not recorded for aspects.

packages/transformer/src/ECSqlReaderAsyncIterableIteratorAdapter.ts
```diff
@@ -0,0 +1,40 @@
...
+    return {
+      done,
+      value: this._ecSqlReader.current,
+    };
+  }
+}
+
+/**
+ * Wraps ECSqlReader with ECSqlReaderAsyncIterableIteratorAdapter if it's needed.
+ * @param ecSqlReader ECSqlReader isntance from itwin 3.x or 4.x version
```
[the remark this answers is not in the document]
mindaugasdirg: Marked

packages/transformer/src/ExportElementAspectsWithElementsStrategy.ts
```diff
@@ -0,0 +1,52 @@
...
+      ._queryAspects(elementId, ElementMultiAspect.classFullName, this._excludedElementAspectClassFullNames)
+      .filter((a) => this.shouldExportElementAspect(a));
+
+    if (multiAspects.length > 0) {
+      this.handler.onExportElementMultiAspects(multiAspects);
+      return this.handler.trackProgress();
+    }
+  }
+
+  public async exportAllElementAspects(): Promise<void> { return Promise.resolve(); }
```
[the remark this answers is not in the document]
mindaugasdirg: Good catch. I added empty implementations in the base class and now child classes can implement only the methods it actually cares about and base classes no longer uses `return Promise.resolve()`

packages/transformer/src/IModelExporter.ts
mindaugasdirg: Yeah, enum was unnecessary here. However, instead of passing constructor I opted to just pass the created strategy object itself. One of my intentions was to allow strategies to support different handler interfaces so that it would be future proof with contract changes. That way there was an option to change aspect exporting methods in `IModelExportHandler` and leave the old strategies unaffected until they are removed. This would allow experimenting with exporting ElementMultiAspects one by one or exporting ElementUniqueAspects and ElementMultiAspects through one method like `shouldExportElementAspect` works with both or different interface changes. So I moved handler creation to strategy itself where each strategy can map handler methods correctly.

packages/transformer/src/IModelExporter.ts
mindaugasdirg: This was only required so that IModelTransformer could set `ElementAspectExportStrategy` based on what parameter it received, but is no longer needed because of your other comment.

packages/transformer/src/IModelTransformer.ts
mindaugasdirg: Yeah, that looks better. It allowed to remove the new option from `IModelTransformer`, remove aspect export strategies enum and `setExportElementAspectsStrategy` from the API of `IModelExporter`.

packages/transformer/src/DetachedExportElementAspectsStrategy.ts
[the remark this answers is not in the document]
MichaelBelousov: fair enough considering the new query

packages/transformer/src/IModelTransformer.ts
mindaugasdirg: Hardly, the issue is that detached aspect exporting will also export aspects for elements which are not exported. This check is for checking if aspect's element was actually exported. The only way I see that this could be moved to `IModelExporter` if it would track which elements were not exported and that list would be used to check if aspect should be exported in detached aspect processing. I'll try to look if there is another approach to this problem.

packages/transformer/src/test/standalone/IModelTransformer.test.ts
mindaugasdirg: Changed it

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
mindaugasdirg: Changed it.

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
mindaugasdirg: Changed it

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
mindaugasdirg: Somehow forgot to commit part of the test code which inserts additional aspect. The test should be good now.

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
mindaugasdirg: Removed `Promise.resolve()` and refactored method to not force async version.

packages/transformer/src/ExportElementAspectsWithElementsStrategy.ts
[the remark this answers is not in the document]
MichaelBelousov: I would prefer the base class doesn't just have no implementation but instead has an optional property/method:

```ts
abstract class Base {
  exportAllElementAspects?(): Promise<void>;
}

/// supports this, property is defined
class Derived extends Base {
  public exportAllElementAspects = async () => {
  }
}

/// doesn't support this (property is undefined)
class NonImplementingDerived extends Base {};
```

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: yes we would need to know the list of exported elements since shouldExportElement is not sufficient if they call process/exportElement themselves... I think I'm OK with what you have here now, so long as you add a comment explaining this issue.

packages/transformer/src/DetachedExportElementAspectsStrategy.ts
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Detached ElementAspect export strategy for [[IModelExporter]].
+ * This strategy exports all ElementAspects separately from the Elements that own them.
+ * @internal
+ */
+export class DetachedExportElementAspectsStrategy extends ExportElementAspectsStrategy<DetachedElementAspectsHandler> {
+  public override registerHandler(handler: () => IModelExportHandler, trackProgress: () => Promise<void>) {
+    super.registerHandler(handler, trackProgress);
+
+    this.handler.onExportElementUniqueAspect = (uniqueAspect, isUpdate) => handler().onExportElementUniqueAspect(uniqueAspect, isUpdate);
```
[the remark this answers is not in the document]
mindaugasdirg: handler callback here wasn't a constructor, but a getter which can return the same instance on every callback. Either way, my intention was to be able to support different element aspect strategy processing handlers so that it would be possible to change the interface and not adjust the old strategies for it, but it creates the problem on how to handle custom handler creation where either `IModelExporter` has to know each strategy and how to register its handler or must call some `registerHandler` method for strategy and provide a lazy way to retrieve `IModelExportHandler` and each handler class would register its portion of handler interface. Neither solution seems clean enough and we don't actually need it now. So I I removed custom aspect handler interfaces and if it will be needed we will be able to add it later when it's actually needed

packages/transformer/src/ExportElementAspectsStrategy.ts
```diff
@@ -0,0 +1,83 @@
...
+      throw new Error("Registered handler getter didn't return handler");
+    }
+    return this._handler;
+  }
+
+  public constructor(sourceDb: IModelDb) {
+    this.sourceDb = sourceDb;
+  }
+
+  public async exportElementAspectsForElement(_elementId: Id64String): Promise<void> { }
```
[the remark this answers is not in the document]
mindaugasdirg: Changed it

packages/transformer/src/ExportElementAspectsStrategy.ts
```diff
@@ -0,0 +1,83 @@
...
+      if (aspect instanceof excludedElementAspectClass) {
+        Logger.logInfo(loggerCategory, `Excluded ElementAspect by class: ${aspect.classFullName}`);
+        return false;
+      }
+    }
+    // ElementAspect has passed standard exclusion rules, now give handler a chance to accept/reject
+    return this.handler.shouldExportElementAspect(aspect);
+  }
+
+  public registerHandler(handler: () => IModelExportHandler, trackProgress: () => Promise<void>) {
```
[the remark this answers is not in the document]
mindaugasdirg: It's not a constructor, but a getter. Also removed this method and the whole explanation is under another comment.

packages/transformer/src/IModelTransformer.ts
mindaugasdirg: Added comment, and also specified under `DetachedExportElementAspectsStrategy` that it will export all aspects regardless if the element was exported or not and that it needs to be handled by the consuming classes.

packages/transformer/src/IModelExporter.ts
```diff
@@ -675,42 +682,10 @@ export class IModelExporter {
...
-      ._queryAspects(elementId, ElementMultiAspect.classFullName, this._excludedElementAspectClassFullNames)
-      .filter((a) => this.shouldExportElementAspect(a));
-
-    if (multiAspects.length > 0) {
-      this.handler.onExportElementMultiAspects(multiAspects);
-      return this.trackProgress();
-    }
+  /** Exports all aspects present in the iModel.
+   */
+  public async exportAllAspects(): Promise<void> {
```
MichaelBelousov, line 687: this adds a very confusing new beta API, that by default doesn't do anything.
I am not sure how best to redesign the API, maybe as `onAllAspectsReady`.

For now, I would prefer we make it a private method that the IModelTransformer class is aware of and uses.

mindaugasdirg: Changed it as you wanted.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1684,6 +1691,7 @@ export class IModelTransformer extends IModelExportHandler {
     await this.initialize(options);
     await this.exporter.exportChanges(options);
     await this.processDeferredElements(); // eslint-disable-line deprecation/deprecation
+    await (this.exporter as any).exportAllAspects();
```
MichaelBelousov, line 1694: typescript allows you to type-safely access private members using bracket syntax:
```ts
await this.exporter["exportAllAspects"();
```
Prefer this since it is type checked. You may need to tell the linter to shut up about the property access syntax, not sure if I merged the test branch where I configured the linter to allow it in the private/protected property case.

mindaugasdirg: Thanks, forgot that this can be used to access properties.

MichaelBelousov (approved):

PR #117, opened 2023-09-08
Reverse synchronize deletes of elements with random ExternalSourceAspects

ViliusRuskys:
see #118 for description


MichaelBelousov:
I had to adjust the ESA scope to not conflict with the targetScopeElement, but otherwise old and new test are passing under the fix. Merging and publishing prerelease

PR #123, opened 2023-09-25
Fix aspect queries when class name is reserved SQLite keyword

mindaugasdirg:
This PR fixes a problem where aspect class has SQLite reserved keyword as its name. If EC class name has SQLite reserved keyword as its name, the class name must be quoted `SELECT * FROM schema.[SQLiteReservedKeyword]`

### DIFF
--- a/packages/transformer/src/DetachedExportElementAspectsStrategy.ts
+++ b/packages/transformer/src/DetachedExportElementAspectsStrategy.ts
@@ -65,7 +65,7 @@ export class DetachedExportElementAspectsStrategy extends ExportElementAspectsSt
   }
 
   private async *queryAspects<T extends ElementAspect>(baseElementAspectClassFullName: string) {
-    const aspectClassNameIdMap = new Map<Id64String, { schemaName: string; className: string; }>();
+    const aspectClassNameIdMap = new Map<Id64String, { schemaName: string, className: string }>();
 
     const optimizesAspectClassesSql = `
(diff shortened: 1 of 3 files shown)


packages/transformer/src/DetachedExportElementAspectsStrategy.ts
```diff
@@ -84,7 +84,7 @@ export class DetachedExportElementAspectsStrategy extends ExportElementAspectsSt
       if(this.excludedElementAspectClassFullNames.has(className))
         continue;
 
-      const getAspectPropsSql = `SELECT * FROM ${className} WHERE ECClassId = :classId ORDER BY Element.Id`;
+      const getAspectPropsSql = `SELECT * FROM ${className.replace(/:(.+)/, ":[$1]")} WHERE ECClassId = :classId ORDER BY Element.Id`;
```
MichaelBelousov, line 87: how about instead you use `ec_className` in the above query twice to get the className and schemaName separately, and then without a string replace just do:
```ts
`SELECT * FROM [${schemaName}].[${className}] WHERE ...`
```
Since you can have the exact same issue with a poorly named schema

MichaelBelousov: also might as well add that case to the test

MichaelBelousov (changes requested):

packages/transformer/src/DetachedExportElementAspectsStrategy.ts
mindaugasdirg: Updated it. It should be a much rarer case, but good catch that we should also handle that.

PR #124, opened 2023-10-24
Handle duplicate code values when updating elements

ViliusRuskys:
This change allows processing of changes where Element CodeValues have been switched with one another

### DIFF
--- a/packages/transformer/src/IModelImporter.ts
+++ b/packages/transformer/src/IModelImporter.ts
@@ -108,7 +108,6 @@ export class IModelImporter implements Required<IModelImportOptions> {
    * A guid string that is used to prefix duplicate code values when performing element updates.
    * The duplicate code value prefixing solves an issue when an element update fails because of a duplicate Element CodeValue in cases where
    * the duplicate target element is set to be deleted in the future, or when elements are switching code values with one another and we need a temporary code value.
-   * Using NULL code values as an alternative is not valid because definition elements cannot have NULL code values.
    * To remove prefixes call [[IModelImporter.resolveDuplicateCodeValues]].
    *
    * @returns undefined if no duplicates were detected.
@@ -238,6 +237,7 @@ export class IModelImporter implements Required<IModelImportOptions> {
           this.onUpdateElement(elementProps);
         } catch(err) {
           if (this.handleElementCodeDuplicates && (err as IModelError).errorNumber === IModelStatus.DuplicateCode) {
+            // Using NULL code values as an alternative is not valid because definition elements cannot have NULL code values.
             this.prefixCode(elementProps);
             this.onUpdateElement(elementProps);
           } else {
--- a/packages/transformer/src/IModelImporter.ts
+++ b/packages/transformer/src/IModelImporter.ts
@@ -5,13 +5,13 @@
 /** @packageDocumentation
  * @module iModels
  */
-import { CompressedId64Set, Id64, Id64String, IModelStatus, Logger } from "@itwin/core-bentley";
+import { CompressedId64Set, DbResult, Guid, GuidString, Id64, Id64String, IModelStatus, Logger } from "@itwin/core-bentley";
 import {
   AxisAlignedBox3d, Base64EncodedString, ElementAspectProps, ElementProps, EntityProps, IModel, IModelError, ModelProps, PrimitiveTypeCode,
   PropertyMetaData, RelatedElement, SubCategoryProps,
 } from "@itwin/core-common";
 import { TransformerLoggerCategory } from "./TransformerLoggerCategory";
-import { ElementAspect, ElementMultiAspect, Entity, IModelDb, Relationship, RelationshipProps, SourceAndTarget, SubCategory } from "@itwin/core-backend";
+import { ECSqlStatement, Element, ElementAspect, ElementMultiAspect, Entity, IModelDb, Relationship, RelationshipProps, SourceAndTarget, SubCategory } from "@itwin/core-backend";
 import type { IModelTransformOptions } from "./IModelTransformer";
 import * as assert from "assert";
 import { deleteElementTreeCascade } from "./ElementCascadingDeleter";
@@ -44,8 +44,8 @@ export interface IModelImportOptions {
    * @default false
    */
   simplifyElementGeometry?: boolean;
-  /** If 'true', try to handle conflicting code value changes by using temporary NULL values.
-   * Code values can be resolved by calling [[IModelImporter.resolveDuplicateCodeValues]].
+  /** If 'true', try to handle duplicate code value changes by prefixing them with a guid value.
+   * Prefixes can be removed by calling [[IModelImporter.resolveDuplicateCodeValues]].
    * @default false
    */
   handleElementCodeDuplicates?: boolean;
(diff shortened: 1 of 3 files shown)


packages/transformer/src/IModelImporter.ts
```diff
@@ -597,9 +638,28 @@ export class IModelImporter implements Required<IModelImportOptions> {
...
+    for (const [elementId, codeValue] of this._duplicateCodeValueMap) {
+      const element = this.targetDb.elements.getElement(elementId);
+      element.code.value = codeValue;
+      element.update();
+    }
+  }
+
+  private trackCodeValueDuplicate(targetElementId: Id64String, codeValue: string) {
+    if (!this._duplicateCodeValueMap)
+      this._duplicateCodeValueMap = new Map<Id64String, string> ();
```
MichaelBelousov, line 659: why make this possibly undefined? An empty map is cheap

packages/transformer/src/IModelImporter.ts
```diff
@@ -597,9 +638,28 @@ export class IModelImporter implements Required<IModelImportOptions> {
       options: this.options,
       targetDbId: this.targetDb.iModelId || this.targetDb.nativeDb.getFilePath(),
       doNotUpdateElementIds: CompressedId64Set.compressSet(this.doNotUpdateElementIds),
+      duplicateCodeValueMap: this._duplicateCodeValueMap ? Object.fromEntries(this._duplicateCodeValueMap) : undefined,
       additionalState: this.getAdditionalStateJson(),
     };
   }
+
+  public resolveDuplicateCodeValues(): void {
```
MichaelBelousov, line 646: why public

packages/transformer/src/IModelImporter.ts
```diff
@@ -207,7 +235,19 @@ export class IModelImporter implements Required<IModelImportOptions> {
       }
     } else {
       if (undefined !== elementProps.id) {
-        this.onUpdateElement(elementProps);
+        try {
+          this.onUpdateElement(elementProps);
+        } catch(err) {
```
MichaelBelousov, line 240: Can we hoist this catch up into the calling function in IModelTransformer, and move the entire duplicate code map into the transformer?

That way it should be easier to keep that map as private, it really shouldn't be public

packages/transformer/src/IModelImporter.ts
```diff
@@ -98,6 +103,28 @@ export class IModelImporter implements Required<IModelImportOptions> {
...
+  /**
+   * A map of conflicting element code values.
+   * In cases where updating an element's code value results in a codeValue conflict, the element's code value will instead be updated to a random guid.
+   * The actual codeValue will be stored in this ElementId->CodeValue map.
+   * This is needed in cases where the duplicate target element is set to be deleted in the future, or when elements are switching code values with one another and we need a temporary code value.
+   * To resolve code values to their intended values call [[IModelImporter.resolveDuplicateCodeValues]].
+   *
+   * @returns undefined if no duplicates were detected.
+   */
+  public get duplicateCodeValueMap(): Map<Id64String, string> | undefined {
```
MichaelBelousov, line 117: I would refrain from making this public. See comment on the `catch` below
(same with the getters/setters which I also think are unnecessary)

packages/transformer/src/IModelTransformer.ts
```diff
@@ -180,6 +180,15 @@ export interface IModelTransformOptions {
...
   ignoreMissingChangesetsInSynchronizations?: boolean;
+
+  /** If 'true', try to handle conflicting code value changes by using temporary guid values.
+   * This solves an issue when an element update fails because of a duplicate Element CodeValue in cases where the duplicate target element is set to be deleted in the future,
+   * or when elements are switching code values with one another and we need a temporary code value.
+   * Code values will be resolved to their intended values during [[finalizeTransformation]].
+   *
+   * @default false
+   */
+  handleElementCodeDuplicates?: boolean;
```
MichaelBelousov, line 191: why put it behind an option? I don't remember if I suggested that... the current implementation doesn't seem like it needs that.

packages/transformer/src/IModelImporter.ts
ViliusRuskys: I don't think hoisting it up would be correct because `importElement` might call `onInsertElement`. We need to react to `DuplicateCode` error only when it comes from `onUpdateElement`. It might hide valid duplicate code cases and these will be impossible to resolve after the transformation.

I made the map public in case the derived transformer classes needed the temporary state of the element, but since there are no such cases yet I can make it private.

packages/transformer/src/IModelImporter.ts
ViliusRuskys: removed nullable attribute

packages/transformer/src/IModelImporter.ts
ViliusRuskys: We need a way to resolve all conflicts after all imports are done. `IModelTransformer.finalizeTransformation` is the place I do this.
Since we removed the option to enable this handling I made this private and exposed a more "generic" method `finalize()`.

packages/transformer/src/IModelImporter.ts
ViliusRuskys: changed everything to private

packages/transformer/src/IModelTransformer.ts
ViliusRuskys: removed the option

packages/transformer/src/IModelImporter.ts
MichaelBelousov: I now noticed you've added `importer.finalize`, I think that is a much better way of doing it :) that looks good to me. Making the map private is sufficient then, thanks!

packages/transformer/src/IModelImporter.ts
MichaelBelousov: this is now private

packages/transformer/src/IModelImporter.ts
```diff
@@ -127,6 +136,7 @@ export class IModelImporter implements Required<IModelImportOptions> {
       preserveElementIdsForFiltering: options?.preserveElementIdsForFiltering ?? false,
       simplifyElementGeometry: options?.simplifyElementGeometry ?? false,
     };
+    this._duplicateCodeValueMap = new Map<Id64String, string> ();
```
MichaelBelousov, line 139: nit: (no need to fix) initialize this map in the property definition itself on line 108

packages/transformer/src/IModelImporter.ts
```diff
@@ -597,9 +620,28 @@ export class IModelImporter implements Required<IModelImportOptions> {
...
+      element.code.value = codeValue;
+      element.update();
+    }
+  }
+
+  /**
+   * Needs to be called to perform necessary cleanup operations.
+   * By not calling `finalize` there is a risk that data imported into targetDb will not be as expected.
+   *
+   * @internal
```
MichaelBelousov, line 640: This should probably not be internal... users of just `IModelImporter` will need it. Maybe note that this doesn't need to be called if using the transformer to manager the importer.

MichaelBelousov (approved):

packages/transformer/src/IModelImporter.ts
ViliusRuskys: removed internal attribute.
Also now clearing the map in `resolveDuplicateCodeValues` , so re-invoking the method will effectvely be a noop.

PR #126, opened 2023-11-13
Added API for querying provenance info for derived transformer classes

ViliusRuskys:
Added methods that allow derived transformer classes to query synchronization version and provenance Scope aspect.

- Federation guid branch is not backwards compatible when it comes to supporting old branches. Exposing api to access scope aspect is necessary so that derived classes can fake the synchronization version themselves.
- Exporting TargetScopeProvenanceJsonProps for derived classes to make sense of 'Scope' aspect's jsonProps.
- explicitly calling `initialize` will throw errors if scope provenance is not initialized in change processing workflow. Exposed `initScopeProvenance` to derived classes.

### DIFF
--- a/packages/transformer/src/IModelTransformer.ts
+++ b/packages/transformer/src/IModelTransformer.ts
@@ -645,13 +645,13 @@ export class IModelTransformer extends IModelExportHandler {
    */
   protected tryGetProvenanceScopeAspect(): ExternalSourceAspect | undefined {
     const sql = `
-    SELECT ECInstanceId
-    FROM ${ExternalSourceAspect.classFullName}
-    WHERE Scope.Id = :scopeId
-      AND Kind = :kind
-      AND Element.Id = :elementId
-      AND Identifier = :identifier
-    LIMIT 1`;
+      SELECT ECInstanceId
+      FROM ${ExternalSourceAspect.classFullName}
+      WHERE Scope.Id = :scopeId
+        AND Kind = :kind
+        AND Element.Id = :elementId
+        AND Identifier = :identifier
+      LIMIT 1`;
 
     const scopeProvenanceAspectId = this.provenanceDb.withPreparedStatement(sql, (stmt) => {
       stmt.bindId("scopeId", IModel.rootSubjectId);
--- a/packages/transformer/src/IModelTransformer.ts
+++ b/packages/transformer/src/IModelTransformer.ts
@@ -206,6 +206,11 @@ class PartiallyCommittedEntity {
   }
 }
 
+/**
+ * Data type for persisting change version information within provenance Scope ExternalSourceAspect.
+ * Additionally, forward synchronization version is stored in Scope aspect's 'version' field.
+ * @beta
+ */
 export interface TargetScopeProvenanceJsonProps {
   pendingReverseSyncChangesetIndices: number[];
   pendingSyncChangesetIndices: number[];
--- a/packages/transformer/src/IModelTransformer.ts
+++ b/packages/transformer/src/IModelTransformer.ts
@@ -585,8 +585,9 @@ export class IModelTransformer extends IModelExportHandler {
 
   /** the changeset in the scoping element's source version found for this transformation
    * @note: the version depends on whether this is a reverse synchronization or not, as
-   * it is stored separately for both synchronization directions
-   * @note: empty string and -1 for changeset and index if it has never been transformed
+   * it is stored separately for both synchronization directions.
+   * @note: must call [[initScopeProvenance]] before using this property.
+   * @note: empty string and -1 for changeset and index if it has never been transformed or was transformed before federation guid update (pre 1.x).
    */
   private get _synchronizationVersion(): ChangesetIndexAndId {
     if (!this._cachedSynchronizationVersion) {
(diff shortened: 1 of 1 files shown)


packages/transformer/src/IModelTransformer.ts
```diff
@@ -606,13 +612,69 @@ export class IModelTransformer extends IModelExportHandler {
     return this._cachedSynchronizationVersion;
   }
 
+  /** the changeset in the scoping element's source version found for this transformation
+   * @note: the version depends on whether this is a reverse synchronization or not, as
+   * it is stored separately for both synchronization directions.
+   * @note: empty string and -1 for changeset and index if it has never been transformed, or was transformed before federation guid update (pre 1.x).
+   */
+  protected get synchronizationVersion(): ChangesetIndexAndId {
+    if (this._cachedSynchronizationVersion) {
```
MichaelBelousov, line 621: nit:
```sh
if (this._cachedSynchronizationVersion === undefined) {
  //...
  this._cachedSynchronizationVersion = "blah";
}
return this._cachedSychronizationVersion;
```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -606,13 +612,69 @@ export class IModelTransformer extends IModelExportHandler {
...
+
+    const version = this._options.isReverseSynchronization
+      ? (JSON.parse(provenanceScopeAspect.jsonProperties ?? "{}") as TargetScopeProvenanceJsonProps).reverseSyncVersion
+      : provenanceScopeAspect.version;
+    if (!version) {
+      return { index: -1, id: "" }; // previous synchronization was done before fed guid update.
+    }
+
+    const [id, index] = version.split(";");
+    this._cachedSynchronizationVersion = { index: Number(index), id };
```
MichaelBelousov, line 638: throw if `Number.isNaN(Number(index))`

packages/transformer/src/IModelTransformer.ts
```diff
@@ -606,13 +612,69 @@ export class IModelTransformer extends IModelExportHandler {
...
+    return this._cachedSynchronizationVersion; // synchronization version found and cached.
+  }
+
+  /**
+   * @returns provenance scope aspect if it exists in the provenanceDb.
+   * Provenance scope aspect is created and inserted into provenanceDb when [[initScopeProvenance]] is invoked.
+   */
+  protected tryGetProvenanceScopeAspect(): ExternalSourceAspect | undefined {
```
MichaelBelousov, line 646: are there any places we can use this in the transformer today where we need this info?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -606,13 +612,69 @@ export class IModelTransformer extends IModelExportHandler {
...
+    const sql = `
+    SELECT ECInstanceId
+    FROM ${ExternalSourceAspect.classFullName}
+    WHERE Scope.Id = :scopeId
+      AND Kind = :kind
+      AND Element.Id = :elementId
+      AND Identifier = :identifier
+    LIMIT 1
+  `;
```
MichaelBelousov, line 655: fix indent

MichaelBelousov (commented):
a few nits, but otherwise will approve after you test for NaN on parsing just in case

packages/transformer/src/IModelTransformer.ts
ViliusRuskys: Yes. We update this aspect in several cases.
1. If we run `processChanges()` on an old branch (that was created before fedGuid update), we fake the version data on the Scope aspect ourselves in order to make the transformer resolve the correct startChangesetIndex.
2. We need to increment reverse sync version by 1 after a reverse synchronization because transformer doesn't take into account that the changeset will be pushed after the transformation on a source iModel. This is necessary for partial fork support.

MichaelBelousov: I actually meant within the core-transformer.

Do you think you could contribute your implementation of "upgrading" an old branching relationship? Assuming it's not specific to "partial forks"

MichaelBelousov: @nick4598 is going to be writing a test of "upgrading" an old branching relationship and your implementation could be helpful

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: I actually meant indent the entire statement :stuck_out_tongue_closed_eyes:

ViliusRuskys: Oops, I somehow missed that 😝
fixed now

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: It doesn't have to be in this PR but we'd appreciate it, we were planning to add that separately

packages/transformer/src/IModelTransformer.ts
```diff
@@ -606,13 +612,69 @@ export class IModelTransformer extends IModelExportHandler {
...
+    const sql = `
+      SELECT ECInstanceId
+      FROM ${ExternalSourceAspect.classFullName}
+      WHERE Scope.Id = :scopeId
+        AND Kind = :kind
+        AND Element.Id = :elementId
+        AND Identifier = :identifier
+      LIMIT 1`;
+
+    const scopeProvenanceAspectId = this.provenanceDb.withPreparedStatement(sql, (stmt) => {
```
MichaelBelousov, line 656: why don't you just reuse `queryScopeExternalSource`? You should probably at least have this call that to handle the query since they appear the same to me.

ViliusRuskys: Replaced the query with `queryScopeExternalSource`

MichaelBelousov (approved):

PR #130, opened 2023-11-30
Add fix to handle max path limit in windows

mindaugasdirg:
Windows has a max total path limit of 260 and there can be schemas which don't reach the max path segment size, but the full path reaches max total path limit in windows systems.


mindaugasdirg:
I've created duplicate PR (#131) to see that CI passes

PR #132, opened 2023-11-30
Handle max path limit on windows for schema names (backport)

mindaugasdirg:
Backport PR for #130 changes

### DIFF
--- a/packages/transformer/src/IModelTransformer.ts
+++ b/packages/transformer/src/IModelTransformer.ts
@@ -1322,7 +1322,9 @@ export class IModelTransformer extends IModelExportHandler {
     let schemaFileName = schema.name + ext;
     // many file systems have a max file-name/path-segment size of 255, so we workaround that on all systems
     const systemMaxPathSegmentSize = 255;
-    if (schemaFileName.length > systemMaxPathSegmentSize) {
+    // windows usually has a limit for the total path length of 256
+    const windowsMaxPathLimit = 260;
+    if (schemaFileName.length > systemMaxPathSegmentSize || path.join(this._schemaExportDir, schemaFileName).length >= windowsMaxPathLimit) {
       // this name should be well under 255 bytes
       // ( 100 + (Number.MAX_SAFE_INTEGER.toString().length = 16) + (ext.length = 13) ) = 129 which is less than 255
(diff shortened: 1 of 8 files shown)


packages/transformer/src/IModelTransformer.ts
```diff
@@ -1322,7 +1322,9 @@ export class IModelTransformer extends IModelExportHandler {
     let schemaFileName = schema.name + ext;
     // many file systems have a max file-name/path-segment size of 255, so we workaround that on all systems
     const systemMaxPathSegmentSize = 255;
-    if (schemaFileName.length > systemMaxPathSegmentSize) {
+    // windows usually has a limit for the total path length of 256
+    const windowsMaxPathLimit = 260;
```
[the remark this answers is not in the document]
mindaugasdirg: Forgot to update comment. I checked and microsoft documentation states that the path limit is 260. https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

MichaelBelousov (approved):
approving since typescript change fixes linter issues when using typescript 5.2
We will want to bump our `@typescript-eslint` version eventually to be able to use new typescript

You may need to add an `// eslint-disable-next-line no-deprecation` comment to pass tests on some versions... we should consider not linting when doing minor version compatibility tests

MichaelBelousov (approved):

pnpm-lock.yaml
```diff
@@ -1,253 +1,344 @@
-lockfileVersion: 5.4
+lockfileVersion: '6.0'
```
MichaelBelousov, line 1: you installed with pnpm 8. This repository uses pnpm 7. Please do:
```sh
pnpm dlx pnpm@^7 install
# or 
npx pnpm@^7 install
```

MichaelBelousov (changes requested):

pnpm-lock.yaml
mindaugasdirg: Fixed it

MichaelBelousov (commented):
LGTM but let's officially deprecate the resumption APIs if we disable the tests, 1.0 will remove them anyway

MichaelBelousov (approved):

PR #135, opened 2023-12-18
Read changesets directly instead of applying them to determine whats changed

nick4598:
Keeping the PR as draft at the moment, because waiting on a change from Affan to the changesetECAdaptor to not error out when there is no ECClassId and instead fall back to the ecclassid of the root class. 

The above scenario occurs when the element is no longer present in the database (because it was deleted in some future changeset that isn't currently being processed) so it can't be queried for its ecclassid. 

EDIT: In 4.3.3 there is a different issue with classIdPresentInChange. See below PR
https://github.com/iTwin/itwinjs-core/pull/6341


packages/transformer/src/IModelExporter.ts
```diff
@@ -942,82 +927,167 @@ export class ChangedInstanceIds {
...
-  private constructor() { }
+  private codeSpecECClassIds: Set<string>;
+  private modelECClassIds: Set<string>;
+  private elementECClassIds: Set<string>;
+  private aspectECClassIds: Set<string>;
+  private relationshipECClassIds: Set<string>;
+  private _db: IModelDb;
+  public constructor(db: IModelDb) {
+    this._db = db;
+    this.codeSpecECClassIds = new Set<string>();
```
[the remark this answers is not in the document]
nick4598: changed

packages/transformer/src/IModelExporter.ts
```diff
@@ -942,82 +927,167 @@ export class ChangedInstanceIds {
...
+  public static async initialize(opts: ChangedInstanceIdsInitOptions): Promise<ChangedInstanceIds> {
     const iModelId = opts.iModel.iModelId;
     const accessToken = opts.accessToken;
 
-    const changesetRanges = opts.changesetRanges
-      ?? [[
-        // we know startChangeset is defined because of the assert above
-        opts.startChangeset!.index
-          ?? (await IModelHost.hubAccess.queryChangeset({
+    let csFileProps: ChangesetFileProps[];
```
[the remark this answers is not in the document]
nick4598: I gave it a shot, let me know what you think. Also relevant to my changes is this comment you made:

"honestly we might want to refactor this into a private initializeFromChangeOptions function, which takes the options, determines which were passed, and returns the ChangedInstanceIds object if necessary after creating. Then hasChangeData can just be a check as to whether that returned undefined." 

Instead of adding the private initFromChangeOptions, I did below instead:
I made it so that ChangedInstanceIds.initialize can return undefined (which might not be allowed? maybe it is since we're going to 1.x?)
and I just check all cases in there, i.e. if opts.changedinstanceids is defined return early. I also get to check the case where there are none of the 4 (changesetranges, startchangeset, changedinstanceids, csFileProps) and return undefined.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2275,18 +1986,215 @@ export class IModelTransformer extends IModelExportHandler {
...
+   * Finds the corresponding target entity for any deleted source entities and remaps the sourceId to the targetId.
+   * Populates this._hasElementChangedCache with a set of elementIds that have been updated or inserted into the database.
+   * This function returns early if csFileProps is undefined or is of length 0.
+   * @returns void
+   */
+  private processChangesets(): void {
+    this.forEachTrackedElement((sourceElementId: Id64String, targetElementId: Id64String) => {
+      this.context.remapElement(sourceElementId, targetElementId);
+    });
+    if (this._csFileProps === undefined || this._csFileProps.length === 0)
```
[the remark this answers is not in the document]
nick4598: from the perspective of the processChangesets function, initialized but without data and not initialized are the same thing, no?
csFileProps can be of length 0 if sourceChangeDataState = "no-changes".
csFileProps is undefined when sourceChangeDataState is 'unconnected' state, but I'm not really sure what that means exactly other than standalonedb? What work can / does the transformer do when the sourcechangedatastate is unconnected / no-changes?.

packages/transformer/src/IModelExporter.ts
```diff
@@ -264,12 +277,11 @@ export class IModelExporter {
...
    */
   public async initialize(options: ExporterInitOptions): Promise<void> {
-    const hasChangeData = options.startChangeset || options.changesetRanges || options.changedInstanceIds;
+    const hasChangeData = "csFileProps" in options || "startChangeset" in options || "changesetRanges" in options || "changedInstanceIds" in options;
     if (this._sourceDbChanges || !this.sourceDb.isBriefcaseDb() || !hasChangeData)
       return;
 
-    this._sourceDbChanges = options.changedInstanceIds
-      ?? await ChangedInstanceIds.initialize({ iModel: this.sourceDb, ...options });
+    this._sourceDbChanges = "changedInstanceIds" in options ? options.changedInstanceIds : await ChangedInstanceIds.initialize({ iModel: this.sourceDb, ...options });
```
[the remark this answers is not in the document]
MichaelBelousov: rather than returning undefined from `initialize` (which perhaps the name should be deprecated and it be renamed to `create`), we should return an empty ChangedInstanceIds. Either the caller should resolve if there were 0 changesets, or we should throw an error on 0 changesets and let the caller catch that... I think. Can discuss when you're back

packages/transformer/src/IModelExporter.ts
```diff
@@ -942,82 +922,155 @@ export class ChangedInstanceIds {
...
+        changeset: { id: startChangeset.id ?? opts.iModel.changeset.id },
+        accessToken,
+      })).index,
+    opts.iModel.changeset.index
+      ?? (await IModelHost.hubAccess.queryChangeset({
+          iModelId,
+          changeset: { id: opts.iModel.changeset.id },
+          accessToken,
+        })).index,
+      ]] : "changesetRanges" in opts ? opts.changesetRanges : undefined;
```
MichaelBelousov, line 1037: put the second ternary on another line

packages/transformer/src/IModelExporter.ts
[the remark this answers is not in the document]
MichaelBelousov: your redo LGTM

packages/transformer/src/test/IModelTransformerUtils.ts
```diff
@@ -280,6 +285,7 @@ export async function assertIdentityTransformation(
         // known cases for the prop expecting to have been changed by the transformation under normal circumstances
         // - federation guid will be generated if it didn't exist
         // - jsonProperties may include remapped ids
+        // FIXME<Mike>: Doesn't this allow propChanges to ANY props if sourceElem.federationGuid is undefined?
         const propChangesAllowed = allowPropChange?.(sourceElem, targetElem, propName)
           ?? (sourceElem.federationGuid === undefined || propName === "jsonProperties");
```
MichaelBelousov, line 290: you can remove the FIXME after you change this to 
```suggestion
          ?? ((propName === "federationGuid" && sourceElem.federationGuid === undefined) || propName === "jsonProperties");
```

packages/transformer/src/test/IModelTransformerUtils.ts
```diff
@@ -302,6 +308,10 @@ export async function assertIdentityTransformation(
             mappedRelationTargetInTargetId
           );
         } else if (!propChangesAllowed) {
+          if (ignoreFedGuidsOnAlwaysPresentElementIds && propName === "federationGuid" && alwaysPresentElementIds.has(sourceElemId)) {
+            // Skip comparing this prop if its the fedguid and its one of the always present element ids and ignoreFedGuidsOnAlwaysPresentElementIds is true.
```
MichaelBelousov, line 312:
```suggestion
            // Skip comparing this prop if its the fedguid and its one of the root elements and ignoreFedGuidsOnAlwaysPresentElementIds is true.
```

MichaelBelousov: actually remove this if and add this condition to the `propChangesAllowed` boolean variable

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
MichaelBelousov: fair enough, resolving

packages/transformer/src/test/IModelTransformerUtils.ts
```diff
@@ -383,7 +393,14 @@ export async function assertIdentityTransformation(
...
       const targetAspectId = remapAspect(sourceAspectId);
-      expect(targetAspectId).not.to.equal(Id64.invalid);
+      const collectAllAspects = (elemId: string) => {
+        let aspects = "";
+        for (const targetElemAspect of targetDb.elements.getAspects(elemId)) {
+          aspects += `\t${JSON.stringify(targetElemAspect)}\n\n`;
+        }
+        return aspects;
+      };
+      expect(targetAspectId, `Expected sourceAspect:\n\t ${JSON.stringify(sourceAspect)}\n on sourceElement:\n\t ${JSON.stringify(sourceElem)}\n with targetElement:\n\t ${JSON.stringify(targetElem)}\n to have a corresponding targetAspectId that wasn't invalid.\n targetElement's aspects:\n${collectAllAspects(targetElemId)}`).not.to.equal(Id64.invalid);
```
MichaelBelousov, line 403: nit: would prefer we use a
```js
[
  string1,
  `string2 ${var}`,
  "string3",
  JSON.stringify(sourceAspect)
].join("\n")
```
approach here. Also multiline `+` for string concat is good too

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -1085,13 +1085,22 @@ describe("IModelTransformer", () => {
...
   });
 
-  /** gets a mapping of element ids to their invariant content */
+  /** gets a mapping of element ids to their content ignoring or removing variance that is expected when transforming */
   async function getAllElementsInvariants(db: IModelDb, filterPredicate?: (element: Element) => boolean) {
+    // The set of element Ids where the fed guid should be ignored (since it can change between transforms).
+    const ignoreFedGuidElementIds = new Set<Id64String>([
+      IModel.rootSubjectId,
+      IModel.dictionaryId,
+      "0xe",
```
MichaelBelousov, line 1094: add a comment saying this is the reality data container (check, I forget what it is)

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -609,19 +609,13 @@ describe("IModelTransformerHub", () => {
             const aspects =
               [...branch.db.queryEntityIds({ from: "BisCore.ExternalSourceAspect" })]
               .map((aspectId) => branch.db.elements.getAspect(aspectId).toJSON()) as ExternalSourceAspectProps[];
+              // FIXME: It seems as if deep subsetEqual requires that the array of elements be in the same order between expected and actual.
```
MichaelBelousov, line 612: I think we provide the `.deep.subsetEqual` implementation as a custom matcher. We can probably support what you want by rewriting it to compare arrays using `.includes`. Or since that's n^2 it could load one into `Set` and do a linear `.has` check

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -1443,6 +1437,29 @@ describe("IModelTransformerHub", () => {
     expect(count(sourceDb, PhysicalModel.classFullName)).to.equal(1);
     expect(count(targetDb, PhysicalModel.classFullName)).to.equal(1);
 
+    sourceDb.elements.deleteElement(secondCopyOfSubjectId);
+    sourceDb.saveChanges();
+    await sourceDb.pushChanges({ description: "deleted the second copy of the subject"});
+    const startChangeset = sourceDb.changeset;
```
MichaelBelousov, line 1443: add a comment like `// readd the subject in a separate changeset`

packages/transformer/src/IModelExporter.ts
```diff
@@ -942,82 +927,167 @@ export class ChangedInstanceIds {
...
+      }
+    });
+    for (const csFile of csFileProps) {
+      const csReader = SqliteChangesetReader.openFile({fileName: csFile.pathname, db: opts.iModel, disableSchemaCheck: true});
+      const csAdaptor = new ChangesetECAdaptor(csReader);
+      const ecChangeUnifier = new PartialECChangeUnifier();
+      while (csAdaptor.step()) {
+        ecChangeUnifier.appendFrom(csAdaptor);
+      }
+      const changes: ChangedECInstance[] = Array.from(ecChangeUnifier.instances);
```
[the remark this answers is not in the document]
nick4598: changed

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
nick4598: I'm looking at this some more and calling Object.keys on an array gives the keys as indices (makes sense). Thats why they have to be in the same order currently. Do your suggestions still work in this case? Since advancedDeepEqual is recursive (basically breaking down each object until its just a simple key value pair, it feels strange to somehow determine which index I'm supposed to be breaking down for comparison when the expected and actual aren't in the same order. I'm fine to leave it as is and maybe call that a TODO: instead of a fixme? 
Code snippet for more context:
```
return (eSize === aSize || !!options.useSubsetEquality) && Object.keys(e).every(
        (keyOfE) =>
          (keyOfE in a || options.considerNonExistingAndUndefinedEqual) &&
          normalizedClassNameProps.includes(keyOfE)
            ? advancedDeepEqual(normalizeClassName(e[keyOfE]), normalizeClassName(a[keyOfE]))
            : advancedDeepEqual(e[keyOfE], a[keyOfE], options)
      );
```

packages/transformer/src/IModelExporter.ts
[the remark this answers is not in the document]
nick4598: Changing to throw instead of return undefined causes lots of tests to fail. Its technically valid to pass no argument isn't it(aka none of the 4: startChangeset, changesetRanges, csFileProps, changedinstanceIds)?. Atleast thats how the type is setup right now. Also some cases like exportAll calls this.initialize() with {}

```
 /** Export all entity instance types from the source iModel.
   * @note [[exportSchemas]] must be called separately.
   */
  public async exportAll(): Promise<void> {
    await this.initialize({});
```

packages/transformer/src/IModelExporter.ts
```diff
@@ -942,82 +927,167 @@ export class ChangedInstanceIds {
...
+
+   private isModel(ecClassId: string) {
+    return this.modelECClassIds.has(ecClassId);
+   }
+
+   private isElement(ecClassId: string) {
+    return this.elementECClassIds.has(ecClassId);
+   }
+
+   public addChange(change: ChangedECInstance): void {
```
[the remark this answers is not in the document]
nick4598: done

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2275,18 +1986,215 @@ export class IModelTransformer extends IModelExportHandler {
...
+      return;
+    const hasElementChangedCache = new Set<string>();
+    const esaNameNormalized = ExternalSourceAspect.classFullName.replace(":", ".");
+
+    const relationshipECClassIdsToSkip = new Set<string>();
+    this.sourceDb.withPreparedStatement(`SELECT ECInstanceId FROM ECDbMeta.ECClassDef where ECInstanceId IS (BisCore.ElementDrivesElement)`, (stmt) => {
```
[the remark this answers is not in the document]
nick4598: done

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2275,18 +1986,215 @@ export class IModelTransformer extends IModelExportHandler {
...
+      const csReader = SqliteChangesetReader.openFile({fileName: csFile.pathname, db: this.sourceDb, disableSchemaCheck: true});
+      const csAdaptor = new ChangesetECAdaptor(csReader);
+      const ecChangeUnifier = new PartialECChangeUnifier();
+      while (csAdaptor.step()) {
+        ecChangeUnifier.appendFrom(csAdaptor);
+      }
+      const changes: ChangedECInstance[] = Array.from(ecChangeUnifier.instances);
+
+      // a map of elementId to ChangedECInstance. the elementId is not the id of the esa itself, but the id that the esa was stored on before the esa's deletion.
+      const esaMap: Map<Id64String, ChangedECInstance> = new Map<Id64String, ChangedECInstance>();
```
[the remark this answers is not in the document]
nick4598: good idea

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2275,18 +1986,215 @@ export class IModelTransformer extends IModelExportHandler {
...
+      for (const change of changes) {
+        if (change.ECClassId !== undefined && relationshipECClassIdsToSkip.has(change.ECClassId))
+          continue; // FIXME: Remove it from the changes array?
```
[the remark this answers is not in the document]
nick4598: removed

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2275,18 +1986,215 @@ export class IModelTransformer extends IModelExportHandler {
...
+    for (const csFile of this._csFileProps) {
+      const csReader = SqliteChangesetReader.openFile({fileName: csFile.pathname, db: this.sourceDb, disableSchemaCheck: true});
+      const csAdaptor = new ChangesetECAdaptor(csReader);
+      const ecChangeUnifier = new PartialECChangeUnifier();
+      while (csAdaptor.step()) {
+        ecChangeUnifier.appendFrom(csAdaptor);
+      }
+      const changes: ChangedECInstance[] = Array.from(ecChangeUnifier.instances);
+
+      // a map of elementId to ChangedECInstance. the elementId is not the id of the esa itself, but the id that the esa was stored on before the esa's deletion.
```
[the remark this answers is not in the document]
nick4598: thanks, i changed it.

packages/transformer/src/IModelExporter.ts
nick4598: done.. I think?

packages/transformer/src/test/IModelTransformerUtils.ts
nick4598: done

packages/transformer/src/test/IModelTransformerUtils.ts
nick4598: done

packages/transformer/src/test/IModelTransformerUtils.ts
nick4598: done

packages/transformer/src/test/standalone/IModelTransformer.test.ts
nick4598: done

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
nick4598: done

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
MichaelBelousov: my suggestion works if you add a check for `Array.isArray` to handle arrays specially from other values where `typeof x === "object"`

packages/transformer/src/IModelExporter.ts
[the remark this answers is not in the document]
MichaelBelousov: iirc there was a default behavior for backwards compat. I suppose we can keep it as optional and keep that behavior, I don't remember it being bad. I'll investigate that just in case tomorrow

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
nick4598: Am I mistaken that your suggestion only works for simple elements in an array? Includes with js objects (which is the case I discovered this issue in) will check that the two objects are exactly the same (same identity?)

MichaelBelousov: discussed this in call

packages/transformer/package.json
```diff
@@ -53,25 +53,25 @@
     "You can find a script to see the latest @itwin/imodel-transformer version for your iTwin.js version in the README"
   ],
   "peerDependencies": {
-    "@itwin/core-backend": "4.3.0-dev.31",
-    "@itwin/core-bentley": "4.3.0-dev.31",
-    "@itwin/core-common": "4.3.0-dev.31",
-    "@itwin/core-geometry": "4.3.0-dev.31",
-    "@itwin/core-quantity": "4.3.0-dev.31",
-    "@itwin/ecschema-metadata": "4.3.0-dev.31"
+    "@itwin/core-backend": "4.3.3",
```
MichaelBelousov, line 56: shouldn't these versions be `^4.3.3`?
Also double check that our lockfile doesn't contain two copies of @itwin/core-backend now

packages/transformer/src/IModelExporter.ts
```diff
@@ -927,26 +927,27 @@ export class ChangedInstanceIds {
   private elementECClassIds = new Set<string>();
   private aspectECClassIds = new Set<string>();
   private relationshipECClassIds = new Set<string>();
+  private _ecClassIdsInitialized = false;
```
MichaelBelousov, line 930: personally I would prefer we just have the sets be `undefined` until initialized, and have a getter with the same name just check if they're undefined.

packages/transformer/src/IModelExporter.ts
```diff
@@ -927,26 +927,27 @@ export class ChangedInstanceIds {
...
+      }
     };
-    addECClassIdsToSet(this.codeSpecECClassIds, "BisCore.CodeSpec");
-    addECClassIdsToSet(this.modelECClassIds, "BisCore.Model");
-    addECClassIdsToSet(this.elementECClassIds, "BisCore.Element");
-    addECClassIdsToSet(this.aspectECClassIds, "BisCore.ElementUniqueAspect");
-    addECClassIdsToSet(this.aspectECClassIds, "BisCore.ElementMultiAspect");
-    addECClassIdsToSet(this.relationshipECClassIds, "BisCore.ElementRefersToElements");
+    const promises = [];
+    promises.push(addECClassIdsToSet(this.codeSpecECClassIds, "BisCore.CodeSpec"));
```
MichaelBelousov, line 943: better more specific name would be like `codeSpecSubclassIds`. Same for all of the super classes.

packages/transformer/src/IModelExporter.ts
```diff
@@ -927,26 +927,27 @@ export class ChangedInstanceIds {
...
+    promises.push(addECClassIdsToSet(this.modelECClassIds, "BisCore.Model"));
+    promises.push(addECClassIdsToSet(this.elementECClassIds, "BisCore.Element"));
+    promises.push(addECClassIdsToSet(this.aspectECClassIds, "BisCore.ElementUniqueAspect"));
+    promises.push(addECClassIdsToSet(this.aspectECClassIds, "BisCore.ElementMultiAspect"));
+    promises.push(addECClassIdsToSet(this.relationshipECClassIds, "BisCore.ElementRefersToElements"));
+    await Promise.all(promises);
```
MichaelBelousov, line 949: nit: not a fan of unnecessary mutable state
```suggestion
    const promises = [
      addECClassIdsToSet(this.codeSpecECClassIds, "BisCore.CodeSpec"),
      addECClassIdsToSet(this.modelECClassIds, "BisCore.Model"),
      addECClassIdsToSet(this.elementECClassIds, "BisCore.Element"),
      addECClassIdsToSet(this.aspectECClassIds, "BisCore.ElementUniqueAspect"),
      addECClassIdsToSet(this.aspectECClassIds, "BisCore.ElementMultiAspect"),
      addECClassIdsToSet(this.relationshipECClassIds, "BisCore.ElementRefersToElements"),
    ];
    await Promise.all(promises);
```

MichaelBelousov (approved):

packages/transformer/package.json
nick4598: I changed both, I only see one hash for @itwin/core-backend. I think we're good.

packages/transformer/src/IModelExporter.ts
nick4598: changed

packages/transformer/src/IModelExporter.ts
nick4598: changed

PR #137, opened 2024-01-02
add test codifying import dynamic schema behavior

MichaelBelousov:
- ECDb requires that importing dynamic schemas use larger versions, even if the version is meaningless, so there is no behavior change we can make to allow importing dynamic schemas
- add a test to demonstrate what happens when you try to import dynamic schemas with compatible and extra classes

### DIFF
--- a/packages/transformer/src/test/standalone/IModelTransformer.test.ts
+++ b/packages/transformer/src/test/standalone/IModelTransformer.test.ts
@@ -2746,6 +2746,101 @@ describe("IModelTransformer", () => {
     targetDb.close();
   });
 
+  it.only("try import dynamic schemas regardless of version", async function () {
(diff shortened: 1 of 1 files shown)


packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -2746,6 +2750,101 @@ describe("IModelTransformer", () => {
     targetDb.close();
   });
 
+  it.only("try import dynamic schemas regardless of version", async function () {
```
MichaelBelousov, line 2753: remove .only

MichaelBelousov:
based on this: https://github.com/iTwin/itwinjs-core/issues/6332

PR #139, opened 2024-01-12
Add prettier as a commit hook, switch to eslint flat config, add some tests surrounding deleting relationships

nick4598:
I believe most/all of the non prettier changes are package.json, pnpm-lock, eslint configs, 
and some tests in imodeltransformerhub.test.ts:
1) it("should be able to handle relationship delete using new relationship provenance method with no fedguids"
1) it("should be able to handle relationship delete using fedguids"
1) it("should be able to handle relationship delete using old relationship provenance method with no fedguids"


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  // "editor.formatOnSave": true, DO NOT TURN THIS ON WITH THE PRETTIER FORMATTER! Known bug with find and replace: https://github.com/prettier/prettier-vscode/issues/3040
```
MichaelBelousov, line 2: fyi I trust you on vscode, I am not going to try it myself
resolve at will

package.json
```diff
@@ -11,6 +11,7 @@
     "test": "pnpm -r test --filter !transformer-performance-tests",
     "docs": "pnpm -r docs",
     "lint": "pnpm -r lint",
+    "prettier:format": "pnpm -r prettier:format",
```
MichaelBelousov, line 14: why not just `format`? What else does prettier do?

MichaelBelousov (commented):
LGTM

MichaelBelousov (approved):
comments are minor so approving pending them

package.json
nick4598: I guess its more explicit this way. Prettier can just output what needs to be changed as opposed to changing it itself. Changing it itself is what prettier:format does.

I guess `format` is an option, but I think I prefer that prettier is mentioned (are there even other formatters? maybe not in which case its pointless)

MichaelBelousov: there are others and eventually we could move to one :P some are even currently working on prettier-compatibility. I would prefer we just say "format" and not reveal the implementation detail that we use prettier

MichaelBelousov: you can resolve this comment (and merge if all comments are resolved) once you do that

PR #140, opened 2024-01-12
Detect synchronization direction from ESA data stored in the provenanceDb

nick4598:


packages/transformer/src/IModelTransformer.ts
```diff
@@ -1270,12 +1348,14 @@ export class IModelTransformer extends IModelExportHandler {
...
+    if (this._isProvenanceInitTransform) return false; // not necessary the first time since there are no deletes to detect
 
-    if (this._options.isReverseSynchronization) return false; // not possible for a reverse synchronization since provenance will be deleted when element is deleted
+    if (this.isReverseSynchronization) return false; // not possible for a reverse synchronization since provenance will be deleted when element is deleted
 
     // FIXME: do any tests fail? if not, consider using @see _isSynchronization
-    if (this._isForwardSynchronization) return false; // not possible for a reverse synchronization since provenance will be deleted when element is deleted
+    // FIXME<NICK> why was this return false before? and why did this work before my changes to determine isReverseSync?
+    // needed to change to true for 'detect element deletes works on children' for this to work.
+    if (this._isForwardSynchronization) return true; // not possible for a reverse synchronization since provenance will be deleted when element is deleted.
```
nick4598, line 1358: What should the comment be here for forward sync "// not possible for a reverse synchronization since provenance will be deleted when element is deleted", since it appears to be copied from the comment for this.isReverseSync.

Something like "not necessary to brute force detect deletions in a forward sync because we can detect all necessary deletes from changeset data" ?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -630,7 +754,8 @@ export class IModelTransformer extends IModelExportHandler {
     args: {
       sourceDb: IModelDb;
       targetDb: IModelDb;
-      // FIXME: deprecate isReverseSync option and instead detect from targetScopeElement provenance
+      // deprecate isReverseSync option and instead detect from targetScopeElement provenance
+      // FIXME<NICK>: how can I deprecate this in a static function? Shouldn't this stay here since it will be passed from transformer?
```
nick4598, line 758: calling this question out

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2170,7 +2251,8 @@ export class IModelTransformer extends IModelExportHandler {
           sourceRelationship,
           targetRelationshipInstanceId
         );
-        aspectProps.id = this.queryScopeExternalSource(aspectProps).aspectId;
+        this.queryScopeExternalSource(this.provenanceDb, aspectProps);
+        // FIXME<NICK>: Why does onExportElement update the aspect, but onExportRelationship doesn't update the aspect?
```
nick4598, line 2253: Calling this question out

packages/transformer/src/IModelTransformer.ts
```diff
@@ -411,18 +413,94 @@ export class IModelTransformer extends IModelExportHandler {
...
   private _isSynchronization = false;
 
-  private get _isReverseSynchronization() {
-    return this._isSynchronization && this._options.isReverseSynchronization;
+  /**
+   * A private variable meant to be set by tests which have an outdated way of setting up transforms. In all synchronizations today we expect to find an ESA in the branch db which describes the master -> branch relationship.
+   * The exception to this is the first transform aka the provenance initializing transform which requires that the master imodel and the branch imodel are identical at the time of provenance initialization.
+   * A couple ofoutdated tests run their first transform providing a source and targetdb that are slightly different which is no longer supported. In order to not remove these tests which are still providing value
+   * this private property on the IModelTransformer exists.
+   * The two tests which currently require this leniency are "Transform source iModel to target iModel" and "should update aspects when processing changes and detachedAspectProcessing is turned on"
```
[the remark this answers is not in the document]
nick4598: I was hoping it would only shrink and we could eventually remove it, but the additional comment probably doesn't help towards that goal, removed.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -386,6 +385,9 @@ export class IModelTransformer extends IModelExportHandler {
   public readonly targetDb: IModelDb;
   /** The IModelTransformContext for this IModelTransformer. */
   public readonly context: IModelCloneContext;
+  private _isReverseSynchronization: boolean | undefined;
```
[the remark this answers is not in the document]
nick4598: Yeah good idea, I kept trying to merge syncType with the public facing isReverseSync / isForwardSync but thats not necessary. Changed it

packages/transformer/src/IModelTransformer.ts
```diff
@@ -411,18 +413,94 @@ export class IModelTransformer extends IModelExportHandler {
...
+   * A private variable meant to be set by tests which have an outdated way of setting up transforms. In all synchronizations today we expect to find an ESA in the branch db which describes the master -> branch relationship.
+   * The exception to this is the first transform aka the provenance initializing transform which requires that the master imodel and the branch imodel are identical at the time of provenance initialization.
+   * A couple ofoutdated tests run their first transform providing a source and targetdb that are slightly different which is no longer supported. In order to not remove these tests which are still providing value
+   * this private property on the IModelTransformer exists.
+   * The two tests which currently require this leniency are "Transform source iModel to target iModel" and "should update aspects when processing changes and detachedAspectProcessing is turned on"
+   */
+  private _allowNoScopingESA = false;
+
+  /**
+   * @returns true for forward sync, false for reverse sync. undefined for no sync.
```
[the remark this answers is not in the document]
nick4598: Changed it

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
nick4598: I also thought about syncDirection, but that seems odd when one of them is "not-sync", not sure.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -411,18 +414,74 @@ export class IModelTransformer extends IModelExportHandler {
...
-    return this._isSynchronization && this._options.isReverseSynchronization;
+  /**
+   * A private variable meant to be set by tests which have an outdated way of setting up transforms. In all synchronizations today we expect to find an ESA in the branch db which describes the master -> branch relationship.
+   * The exception to this is the first transform aka the provenance initializing transform which requires that the master imodel and the branch imodel are identical at the time of provenance initialization.
+   * A couple ofoutdated tests run their first transform providing a source and targetdb that are slightly different which is no longer supported. In order to not remove these tests which are still providing value
+   * this private property on the IModelTransformer exists.
+   */
+  private _allowNoScopingESA = false;
+
+  private determineSyncType(): SyncType {
```
MichaelBelousov, line 425: make this a public static API

packages/transformer/src/IModelTransformer.ts
```diff
@@ -369,6 +368,8 @@ type ChangeDataState =
  */
 export type InitFromExternalSourceAspectsArgs = InitOptions;
 
+type SyncType = "not-sync" | "forward" | "reverse";
```
nick4598, line 371: Do I have to export this type now that determineSyncType is a public static function which returns that type?

packages/transformer/src/IModelTransformer.ts
nick4598: done

packages/transformer/src/IModelTransformer.ts
nick4598: hm nevermind, i made it so determineSyncType returns "forward" | "reverse", even though SyncType also has "not-sync". I feel like not-sync only applies to state determined within the transformer so I don't think it makes sense to return "not-sync" within the static function.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -411,18 +414,144 @@ export class IModelTransformer extends IModelExportHandler {
...
+      LIMIT 1
+    `;
+    dbToQuery.withPreparedStatement(sql, (statement: ECSqlStatement) => {
+      statement.bindId("elementId", aspectProps.element.id);
+      if (aspectProps.scope === undefined) return; // return instead of binding an invalid id
+      statement.bindId("scopeId", aspectProps.scope.id);
+      statement.bindString("kind", aspectProps.kind);
+      statement.bindString("identifier", aspectProps.identifier);
+      if (DbResult.BE_SQLITE_ROW !== statement.step()) return;
+      aspectProps.id = statement.getValue(0).getId();
```
MichaelBelousov, line 458: iirc the original code did this, but let's take this chance to remove mutation of input variables and return a new copy instead

packages/transformer/src/IModelTransformer.ts
```diff
@@ -411,18 +414,144 @@ export class IModelTransformer extends IModelExportHandler {
...
+
+  /**
+   * querys the scopes external source and sets aspectId, version and jsonProperties on the provided aspectProps if available.
+   * @param dbToQuery db to run the query on for scope external source
+   * @param aspectProps aspectProps to search for
+   * @param opts opts containing preserveVersionReceived and getJsonProperties. if preserveVersionReceived is true and an aspect is found with the provided aspectProps,
+   * then the version passed as part of the aspectProps will be kept on the aspectProps.
+   * If false or not provided, the version on the aspectProps will be overwritten with the version from the queried aspect.
+   */
+  public static queryScopeExternalSource(
```
MichaelBelousov, line 436: Rename from `queryScopeExternalSource` to `queryTargetScopeExternalSourceAspect`?

Also can we always query the JSON properties and just ignore them in callers that don't need them? I don't remember why I differentiated it last time I refactored however... if we find there is a reason, let's document it over the `getJsonProperties` option

packages/transformer/src/IModelTransformer.ts
```diff
@@ -411,18 +414,144 @@ export class IModelTransformer extends IModelExportHandler {
...
+   * @throws if no scoping ESA can be found in either the sourceDb or targetDb which describes a master branch relationship between the two databases.
+   * @param sourceDb
+   * @param targetDb
+   * @param targetScopeElementId the element Id defined on the ExternalSourceAspect's scope. @see [[ExternalSourceAspectProps]]
+   * @returns "forward" or "reverse"
+   */
+  public static determineSyncType(
+    sourceDb: IModelDb,
+    targetDb: IModelDb,
+    targetScopeElementId: Id64String
```
MichaelBelousov, line 481: nit:
```suggestion
   * @returns "forward" or "reverse"
   */
  public static determineSyncType(
    sourceDb: IModelDb,
    targetDb: IModelDb,
    /** @see [[IModelTransformOptions.targetScopeElement]] */
    targetScopeElementId: Id64String
```
this is purely stylistic, resolve as you see fit. But imo it's more consistent when there are edges cases like inline parameter types

Also please double check my reference to `IModelTransformOptions.targetScopeElement`, but I think centralizing that definition in the docs is better.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -411,18 +414,144 @@ export class IModelTransformer extends IModelExportHandler {
...
+      element: {
+        id: targetScopeElementId,
+        relClassName: ElementOwnsExternalSourceAspects.classFullName,
+      },
+      scope: { id: IModel.rootSubjectId }, // the root Subject scopes scope elements
+      identifier: sourceDb.iModelId,
+      kind: ExternalSourceAspect.Kind.Scope,
+      jsonProperties: undefined as TargetScopeProvenanceJsonProps | undefined,
+    };
+    // First assume that the targetDb is the provenanceDb/branch (meaning holds the provenance which points back to master)
```
MichaelBelousov, line 496:
```suggestion
    // First check if the targetDb is the branch 
```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -411,18 +414,144 @@ export class IModelTransformer extends IModelExportHandler {
...
+    this.queryScopeExternalSource(targetDb, aspectProps, {
+      getJsonProperties: true,
+    });
+    if (undefined !== aspectProps.id) {
+      return "forward"; // we found an esa assuming targetDb is the provenanceDb/branch so this is a forward sync.
+    }
+
+    // Now assume that the sourceDb is the provenanceDb/branch
```
MichaelBelousov, line 504:
```suggestion
    // Now check if the sourceDb is the branch
```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -411,18 +414,144 @@ export class IModelTransformer extends IModelExportHandler {
...
+    try {
+      return IModelTransformer.determineSyncType(
+        this.sourceDb,
+        this.targetDb,
+        this.targetScopeElementId
+      );
+    } catch (err) {
+      if (
+        err instanceof Error &&
+        err.message === IModelTransformer.noEsaSyncDirectionErrorMessage &&
```
MichaelBelousov, line 532: nit, consider:
```ts
interface SyncDirectionErr extends Error {
  code: "TRANSFORMER-SYNCDIR";
}
const isSyncDirectionErr = (e: Error) => (e as SyncDirectionErr).code === "TRANSFORMER-SYNCDIR";
```
or
```ts
class SyncDirectionError extends Error {}
// and use `instanceof`
```

Note that I consider `instanceof` to be bad practice in javascript because it isn't duck typed and doesn't work if you end up importing multiple versions of one library. Not a major issue but I prefer to not use it in general.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -411,18 +414,144 @@ export class IModelTransformer extends IModelExportHandler {
...
+  }
+
+  public get isForwardSynchronization(): boolean {
+    if (this._syncType === undefined) this._syncType = this.determineSyncType();
+    return this._syncType === "forward";
   }
 
   private _changesetRanges: [number, number][] | undefined = undefined;
 
   /** Set if it can be determined whether this is the first source --> target synchronization. */
```
MichaelBelousov, line 553:
```suggestion
  /**
   * Set if the transformer is being used to perform the provenance initialization step of a fork initialization.
   * In general don't use the transformer for that, prefer [[BranchingProvenanceInitializer.initializeFork]]
   */
```

packages/transformer/src/IModelTransformer.ts
nick4598: I changed it to mutate the input variable intentionally, because you had commented that it was strange we were returning aspectProps. Was the strange part that we used to be returning the below type as opposed to a type of ExternalSourceAspectProps ? 

` private queryScopeExternalSource(
    aspectProps: ExternalSourceAspectProps,
    { getJsonProperties = false } = {}
  ): {
    aspectId?: Id64String;
    version?: string;
    /** stringified json */
    jsonProperties?: string;
  } {`

packages/transformer/src/IModelTransformer.ts
nick4598: `queryTargetScopeExternalSourceAspect`. Is TargetScopeESA different than the scoping ESA? I suspect it is, but just asking since this function is used for more than just the scoping ESA

packages/transformer/src/IModelTransformer.ts
nick4598: I think I'd prefer to keep the lengthier comment, but I did change assume to 'check if'. I just think it gives more context to someone looking at this repo for the first time

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: maybe?
After looking at it again, I'm pretty sure it _used_ to mutate, I changed it to return which i think is better. Idr what I considered strange exactly, but showing me the original comment would help me...
I do think returning a subset of the aspect props is ok in general, makes it more widely reusable

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: yes I think I meant to go back and edit this comment before I posted my review. Let's just drop `Target` and do `queryScopeExternalSourceAspect`? I wanted to add "Target" to align better with the concept of `targetScopeElement`, which is always in the branch (hence "target", but in a reverse sync the branch (also provenance db) is the source)
But I think you're right we use it in other places that makes that less of a good idea

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: I would personally rather something like:
```
//First check if the target is the branch (branch is the @see provenanceDb)
```
I prefer we link back to centralized doc concepts as much as possible

packages/transformer/src/IModelTransformer.ts
nick4598: Okay I'll go back to not mutating then, the original comment was over a call so no records of it.

packages/transformer/src/IModelTransformer.ts
nick4598: Regarding `getJsonProperties`, all tests still pass if we are to remove the `getJsonProperties` option and just always grab the jsonProperties. I guess my only concern is can we be sure that every ESA has jsonProperties? Previously this function was private and is now public static, so I think in theory anyone could look for any type of ESA, and jsonProperties are not required to exist on ESA

packages/transformer/src/IModelTransformer.ts
nick4598: Changed

packages/transformer/src/IModelTransformer.ts
nick4598: Added check for isNull before getting the jsonprops.

packages/transformer/src/IModelTransformer.ts
nick4598: Yeah thats better, changed it.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1049,11 +1051,18 @@ export class IModelTransformer extends IModelExportHandler {
...
           jsonProperties: JSON.stringify(aspectProps.jsonProperties) as any,
         });
+        aspectProps.id = id;
       }
+    } else {
+      aspectProps.id = foundEsaProps.aspectId;
+      aspectProps.version = foundEsaProps.version;
+      aspectProps.jsonProperties = foundEsaProps.jsonProperties
+        ? JSON.parse(foundEsaProps.jsonProperties)
+        : {};
```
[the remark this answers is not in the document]
nick4598: Yeah I think this might be left over and realized how weird this is when I was working on my other PR about handling imodels last synced before the version change. I will fix it up in my next PR if thats alright with you, because in order to support old imodels we probably need to instead setup the pendingreversesyncversion and pendingsyncversion jsonProperties if jsonProperties end up being undefined.

Issue #134, opened 2023-12-11
IModelTransformer is not handling unresolved elements in finalizeTransformation()


JulijaRamoskiene:
In IModelTransformer:finalizeTransformation() unresolved elements are not logged and  forceComplete () is not called for them.

After running transformation I see this warning: 
```code
Warning | core-backend.IModelTransformer | The following elements were never fully resolved:

This indicates that either some references were excluded from the transformation
or the source has dangling references. {}
```

**Additional context**
This issue is happening because BigMap does not have keys () and values() implementation that would correctly collect keys and values.
https://github.com/iTwin/imodel-transformer/blob/main/packages/transformer/src/BigMap.ts

https://github.com/iTwin/imodel-transformer/blob/main/packages/transformer/src/IModelTransformer.ts#L1122C6-L1122C6
private finalizeTransformation() {
    if (this._partiallyCommittedEntities.size > 0) {
      Logger.logWarning(
        loggerCategory,
        [
          "The following elements were never fully resolved:",
          **[...this._partiallyCommittedEntities.keys()].join(","),**
          "This indicates that either some references were excluded from the transformation",
          "or the source has dangling references.",
        ].join("\n")
      );
      **for (const partiallyCommittedElem of this._partiallyCommittedEntities.values()) {**
        partiallyCommittedElem.forceComplete();
      }
    }

MichaelBelousov:
Since you identified the problem, would you be willing to submit a PR fixing this @JulijaRamoskiene?

JulijaRamoskiene:
Not at the moment, as I am a bit occupied with my current task. I will let you know if I have time in the future (to check if someone else started working on it).

PR #144, opened 2024-01-23
Implemented missing methods in BigMap 

JulijaRamoskiene:
Closes https://github.com/iTwin/imodel-transformer/issues/134
- Changed BigMap to implement Map instead of extending it. Implemented missing methods.
- Added BigMap tests.
- Added iModelTransformer test to verify that unresolved references are logged correctly. Moved logger initialization logic to beforeEach(), as it is reconfigured in new test case.

### DIFF
--- a/packages/transformer/src/test/standalone/IModelTransformer.test.ts
+++ b/packages/transformer/src/test/standalone/IModelTransformer.test.ts
@@ -19,7 +19,7 @@ import {
 import * as coreBackendPkgJson from "@itwin/core-backend/package.json";
 import * as ECSchemaMetaData from "@itwin/ecschema-metadata";
 import * as TestUtils from "../TestUtils";
-import { DbResult, Guid, Id64, Id64String, Logger, LogLevel, OpenMode } from "@itwin/core-bentley";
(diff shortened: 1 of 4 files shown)


nick4598 (approved):

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -642,6 +645,56 @@ describe("IModelTransformer", () => {
...
+
+    // Collect expected ids
+    const result = iModelA.queryEntityIds({ from: "BisCore.Element",
+      where: "Model.Id = :rootId AND ECInstanceId NOT IN (:rootId, :excludedId)",
+      bindings: {rootId: IModel.rootSubjectId, excludedId } });
+    const expectedIds = [...result].map ((x) => `e${x}`);
+
+    // Collect actual ids
+    assert.isDefined(unresolvedElementMessage);
+    const actualIds = unresolvedElementMessage!.split(messageStart)[1].split(messageEnd)[0].split(",");
```
MichaelBelousov, line 687: Nit: you could just match a regex:
```ts
/^start.*end$/ms.test(log)
```

MichaelBelousov (approved):
The test could be smaller but it's ok as is

PR #146, opened 2024-01-31
Allow no versioning data if user opts into it

nick4598:
Also documented targetScopeProvenanceJsonProps some more and added a test


packages/transformer/src/test/TestUtils/TimelineTestUtil.ts
```diff
@@ -525,9 +540,10 @@ export async function runTimeline(
 
         target.state = getIModelState(target.db); // update the tracking state
 
-        if (!isForwardSync)
+        if (!isForwardSync && !expectThrow)
```
MichaelBelousov, line 543: apparently I can't suggest through both since a line in the middle didn't change, but why not:

```ts
if (!expectThrow) {
  if (!isForwardSync)
    await saveAndPushChanges(accessToken, source.db, stateMsg);
  await saveAndPushChanges(accessToken, target.db, statMsg);
}
```

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -3255,6 +3255,188 @@ describe("IModelTransformerHub", () => {
     masterSeedDb.close();
   });
 
+  it("should fail older iModels without new versioning behavior unless ignoreNoBranchRelationshipData is true", async () => {
```
MichaelBelousov, line 3258: if I understand correctly
```suggestion
  it("should fail processingChanges on pre-version-tracking forks unless ignoreNoBranchRelationshipData is true", async () => {
```

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -3255,6 +3255,188 @@ describe("IModelTransformerHub", () => {
...
+    let targetScopeProvenanceProps: ExternalSourceAspectProps | undefined;
+    const setIgnoreNoBranchRelationshipData = (
+      transformer: IModelTransformer
+    ) => (transformer["_options"]["ignoreNoBranchRelationshipData"] = true);
+    const timeline: Timeline = [
+      { master: { 1: 1 } },
+      { master: { 2: 2 } },
+      { master: { 3: 1 } },
```
MichaelBelousov, line 3266: is it important that these are separate changesets? why not:
```suggestion
      { master: { 1: 1, 2: 2, 3: 1 } },
```

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -3255,6 +3255,188 @@ describe("IModelTransformerHub", () => {
...
+      },
+      // eslint-disable-next-line @typescript-eslint/no-shadow
+      {
+        assert({ master, branch }) {
+          expect(master.db.changeset.index).to.equal(5);
+          expect(branch.db.changeset.index).to.equal(8);
+          expect(count(master.db, ExternalSourceAspect.classFullName)).to.equal(
+            0
+          );
+          // added because the root was modified
```
MichaelBelousov, line 3384: what was added? How many were added? can you redefine `11` in terms of the amount there were before `+` something?

packages/transformer/src/test/TestUtils/TimelineTestUtil.ts
```diff
@@ -500,12 +504,23 @@ export async function runTimeline(
             accessToken,
             startChangeset: startIndex ? { index: startIndex } : undefined,
           });
+          expect(
+            expectThrow === false || expectThrow === undefined,
```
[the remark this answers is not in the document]
nick4598: I believe I would also have to change the .to.be.true to to.be.false to do this right? Except chai actually seems to only accept 'false' i.e. undefined doesn't mean false.

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
nick4598: I also copied this from another test, but I just added a comment explaining 11 ESAs. Let me know your thoughts

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
nick4598: Just copied from other tests, changed it to 1 changeset.

packages/transformer/src/test/TestUtils/TimelineTestUtil.ts
nick4598: changed

packages/transformer/src/IModelTransformer.ts
```diff
@@ -258,6 +258,16 @@ export interface IModelTransformOptions {
    * @default false
    */
   ignoreMissingChangesetsInSynchronizations?: boolean;
+
+  /**
+   * Do not error out if a scoping ESA @see ExternalSourceAspectProps is found without a version or jsonProperties defined on that scoping ESA.
+   * If true, the version and jsonproperties will be properly set on the scoping ESA @see TargetScopeProvenanceJsonProps after the transformer is complete.
+   * These properties not being defined are a sign that this branching relationship was created with an older version of the transformer, and setting this option to true is not without risk.
+   * @note Depending on the state of the branching relationship at the time of using this option, some data may be lost.
+   * @note It is very unlikely that this option will ever need to be provided.
```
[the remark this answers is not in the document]
nick4598: agreed, changed it to allowNoBranchRelationshipData

packages/transformer/src/test/TestUtils/TimelineTestUtil.ts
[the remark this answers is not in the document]
MichaelBelousov: wait sorry. I meant `!expectThrow`. confused myself :)

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
MichaelBelousov: why not just:
```ts
expect(count(branch.db, "bis.ExternalSourceAspect")).to.equal(
       count(branch.db, "bis.Element"));
```
I think that better indicates it. In general comments like this can become stale easily, it's best to "comment assumptions with code" in tests

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
nick4598: Correct, its a one time operation. Do you think migrate gives off the wrong impression that this is a smooth process? Data loss is definitely possible.
I did decide to add a note saying this should only need to done at most once for a branching relationship.

packages/transformer/src/test/TestUtils/TimelineTestUtil.ts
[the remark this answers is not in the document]
nick4598: I think i'd prefer to keep `expectThrow === false || expectThrow === undefined` because it feels easier to read to me than `!expectThrow`. Let me know if you still want me to change it and I will.

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
nick4598: I agree that seems better, but I also think theres value in me fully explaining the contents of the database at that point in time. Maybe this is the wrong place to do it, but somewhere is better than nowhere imo. I think its one small step towards better understanding of the bis stuff in general

If the number were to change from 11, I'd probably want to know for my own enrichment because maybe that means theres another element? 

I did change the comment around a bit just now, but not significantly imo.

MichaelBelousov: I still think a comment is a future source of staleness.
What do you think about converting your comment into:
```ts
const externalAspectCounts = (db) => db.query(`
  SELECT e.ECInstanceId, COUNT(*) FROM bis.ExternalSourceAspect esa
  JOIN bis.Element e ON e.ECInstanceId=esa.Element.Id
  GROUP BY e.ECInstanceId
`);
  

expect(count(db, "bis.ExternalSourceAspect")).to.be(11);
expect(count(db, "bis.Element)).to.be(10);

for (const [id, aspectCount] of externalAspectCounts(db)) {
  if (id === transformer.targetScopeElement)
    expect(aspectCount).to.equal(2);
  else
    expect(aspectCount).to.equal(1);
}
```

The full db contents really belong not here but elsewhere, earlier in the `it` scope, I assume right after initialization of the db

MichaelBelousov: I see now you wanted more "at that point in time".
really imo the need for this is due to poor diffing support elsewhere

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
MichaelBelousov: I think we decided in chat that `"unsafe-migrate"` would be better, with an explanation

MichaelBelousov (approved):

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
nick4598: The comment being a source of staleness is a good thing imo, we should understand when our database suddenly has 12 elements instead of 11. I removed the comment 

Yeah agree about the diffing support, what I've been doing locally is modifying the count helper function to push rows to an array that I can look at in the debugger. Not sure if you're opposed to me committing that as I think it would be slower. ( I wouldn't add it in this PR just some other time)

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
nick4598: changed to unsafe-migrate

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
MichaelBelousov: feel free to commit it under an environment variable based branch

MichaelBelousov (approved):

nick4598:
### Merge activity

* **Feb 23, 1:04 PM EST**: @nick4598 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/iTwin/imodel-transformer/146).

PR #150, opened 2024-02-15
throw when danglingReferencesBehavior is reject

nick4598:
throw when danglingReferencesBehavior is reject

PR #151, opened 2024-02-15
Support delete detection in test

nick4598:


nick4598:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/iTwin/imodel-transformer/151?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

Current dependencies on/for this PR:
* `federation-guid-optimization`
  * **PR #146** <a href="https://app.graphite.dev/github/pr/iTwin/imodel-transformer/146?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #150** <a href="https://app.graphite.dev/github/pr/iTwin/imodel-transformer/150?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #151** <a href="https://app.graphite.dev/github/pr/iTwin/imodel-transformer/151?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
        * **PR #152** <a href="https://app.graphite.dev/github/pr/iTwin/imodel-transformer/152?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/iTwin/imodel-transformer/151?utm_source=stack-comment).

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -399,39 +405,40 @@ describe("IModelTransformer", () => {
...
+      TransformerExtensiveTestScenario.assertUpdatesInDb(targetDb, true);
 
       assert.equal(
         numTargetRelationships +
           targetImporter.numRelationshipsInserted -
           targetImporter.numRelationshipsDeleted,
         count(targetDb, ElementRefersToElements.classFullName)
       );
-      // FIXME: why?
+      // FIXME<NICK>: I saw that the previous commit which added this FIXME: why changed the test from 3 to 2. But now in my test its producing 2.
```
nick4598, line 437: Calling attn to this comment

packages/transformer/src/test/IModelTransformerUtils.ts
```diff
@@ -982,32 +984,38 @@ export class TransformerExtensiveTestScenario extends TestUtils.ExtensiveTestSce
     TransformerExtensiveTestScenario.assertTargetElement(
       sourceDb,
       targetDb,
-      definitionModelId
+      definitionModelId,
+      expectEsas
```
MichaelBelousov, line 988: `TransformerExtensiveTestScenario.assertTargetElement` is only used within this function. I would recommend just defining it as a local function so it can close over expectEsas and you won't need to add all these new arguments. In fact, you can probably get rid of all but the id parameter if you do that.

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -282,7 +284,9 @@ describe("IModelTransformer", () => {
       targetDb.saveChanges();
       TransformerExtensiveTestScenario.assertTargetDbContents(
         sourceDb,
-        targetDb
+        targetDb,
+        undefined,
+        true
```
MichaelBelousov, line 289: having `undefined` in the middle of parameters is a sign that you should refactor to use an argument object with optional properties. There are only 4 calls to `assertTargetDbContents` so it shouldn't be that bad

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -399,39 +405,40 @@ describe("IModelTransformer", () => {
...
       assert.equal(targetImporter.numModelsUpdated, 0);
       assert.equal(targetImporter.numElementsInserted, 1);
       assert.equal(targetImporter.numElementsUpdated, 33);
-      // FIXME: upgrade this test to use a briefcase so that we can detect element deletes
-      // use the new force old detect deletes behavior flag here
-      // assert.equal(targetImporter.numElementsDeleted, 5);
+      /**
+       * There are 5 elements deleted in TransformerExtensiveTestScenario.updateDb, but only 4 detected.
+       * This is because PhysicalObject6's code is scoped to PhysicalObject5. When PhysicalObject5 is deleted, PhysicalObject6 is also deleted in the superclasses
+       * of the RecordingIModelImporter and therefore can't be detected by the RecordingIModelImporter.
```
MichaelBelousov, line 419: maybe this is a good time to stop using `RecordingIModelImporter`? You could alternatively change the tree deletions (iirc this is the problem) to return the set of deletes somehow for the RecordIModelImporter to process. But I'd rather we not perform an iModel insert for every delete... I think it's awkward and probably brittle.

However you do this, assert that the fifth element is no longer in the iModel, and maybe rename this to `numElementsExplicitlyDeleted` and document on that property that it doesn't include implicit deletions when deleting trees.

packages/transformer/src/test/standalone/IModelTransformer.test.ts
MichaelBelousov: `TargetInformationRecord` elements seem to come from `TestIModelTransformer.initClassRemapping` which remaps all `SourceInformationRecord`s to it. Other tests find 3 of them so I'd be concerned about the 2 as well.
We should check how many `SourceInformationRecords` there are in the source (I assume 3) and see if we deleted any of them.

.gitignore
```diff
@@ -19,6 +19,9 @@ imjs_plugins
 imjs_extensions
 !tools/build
 
+# git-specific
+.git-blame-ignore-revs
```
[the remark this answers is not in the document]
nick4598: Maybe I set this up improperly, but I wanted to ignore my prettier commit so I could see who actually made a change to the code I was looking at.

It seems .git-blame-ignore-revs is a naming convention (not sure how strong of a convention this is), but I set up one of these files in my local transformer repo and I didn't want to force anyone to skip the prettier commit if they didn't want to i.e. by committing the file. Although I guess it seems like I also had to point to this file in my git config, so maybe it doesn't matter..

packages/transformer/src/test/standalone/IModelTransformer.test.ts
nick4598: changed it

packages/transformer/src/test/standalone/IModelTransformer.test.ts
nick4598: I checked and we do delete one, updated the comment

packages/transformer/src/test/IModelTransformerUtils.ts
nick4598: changed

packages/transformer/src/test/standalone/IModelTransformer.test.ts
nick4598: I ended up keeping RecordingIModelImporter as is but, I did change the numElementsDeleted to explicitlyDeleted.

Regarding 'return the set of deletes somehow for the RecordIModelImporter to process', I did have a brief thought about this that any custom importers might run into this issue if they're doing something for each delete, but I'm hoping thats not the case and I'd rather defer this problem for the future / if theres a request or need for it.

.gitignore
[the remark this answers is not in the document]
MichaelBelousov: I think you can use a local one and remove it. Maybe commit it with some comments on how to use it for those interested? It is useful for sure

MichaelBelousov (approved):

PR #153, opened 2024-02-15
save and push changes after transformer is complete

nick4598:

PR #152, opened 2024-02-15
Run lint:fix

nick4598:

PR #154, opened 2024-02-15
rootElementPropagateChanges/skipBranchProvInitializeCs

nick4598:
rootElementPropagateChanges/skipBranchProvInitializeCs

Skip over repositorymodel in the remap case, remove some comments


nick4598:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/iTwin/imodel-transformer/154?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

Current dependencies on/for this PR:
* `federation-guid-optimization`
  * **PR #146** <a href="https://app.graphite.dev/github/pr/iTwin/imodel-transformer/146?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #150** <a href="https://app.graphite.dev/github/pr/iTwin/imodel-transformer/150?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #151** <a href="https://app.graphite.dev/github/pr/iTwin/imodel-transformer/151?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #152** <a href="https://app.graphite.dev/github/pr/iTwin/imodel-transformer/152?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #153** <a href="https://app.graphite.dev/github/pr/iTwin/imodel-transformer/153?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
            * **PR #154** <a href="https://app.graphite.dev/github/pr/iTwin/imodel-transformer/154?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/iTwin/imodel-transformer/154?utm_source=stack-comment).

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1132,11 +1146,11 @@ export class IModelTransformer extends IModelExportHandler {
       ? args.provenanceSourceDb
       : args.provenanceDb;
 
+    // FIXME<NICK>: I removed the where e.ecinstanceId not in (0x1, 0xe, 0x10) here. I wonder if we need to special case it for the propagatechangesoption.
```
nick4598, line 1149: All tests pass with or without the previous WHERE clause. I may have removed this preemptively when I was first trying to support changes to the root elements. 

I'm wondering if in the cases where _options.skipPropagateChangesToRootElements is set to true, 0x1,0xe,0x10 would get filtered out elsewhere in the code anyways.  

I guess thats only true right now because the only way we use forEachTrackedElement in the transformer is to call context.remapElement. Theoretically someone could probably pass a more complex function to forEachTrackedElement that modifies which to me seems like an argument in favor of adding the where clause based off of the truth value of the new skipPropagateChangesToRootElements option. Any thoughts?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -727,6 +717,30 @@ export class IModelTransformer extends IModelExportHandler {
...
+   *  @note This expects that the importer is already set on the transformer.
+   */
+  private validateSharedOptionsMatch() {
+    if (
+      Boolean(this._options.preserveElementIdsForFiltering) !==
+      this.importer.options.preserveElementIdsForFiltering
+    ) {
+      const errMessage =
+        "A custom importer was passed as a target but its 'preserveElementIdsForFiltering' option is out of sync with the transformer's option.";
+      Logger.logError(loggerCategory, errMessage);
```
MichaelBelousov, line 730: don't log what you throw, let the consumer decide if they want to log it

MichaelBelousov: not sure if historically I've followed that enough myself, but I think it's best practice

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: I would add the where clause. We can always remove it if we find issues.
We shouldn't rely on invariants up in the call tree without asserting them, the order may change eventually.
Hopefully a rally on testing will fix that stuff.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1930,12 +1944,24 @@ export class IModelTransformer extends IModelExportHandler {
...
+    if (
+      sourceModel.id === IModel.repositoryModelId &&
+      targetModeledElementId != sourceModel.id
+    ) {
+      /**
+       * The repositoryModel should not be directly imported if its targetModeledElementId is different than the sourceModel.id, since this means the repositoryModel/rootSubject was remapped.
+       * There can only be one repositoryModel per database so we will return early and not import it.
+       */
+      return;
+    }
```
MichaelBelousov, line 1964:
```suggestion
    // there can only be one repositoryModel per database, so ignore the repo model on remapped subjects
    const isRemappedRootSubject = sourceModel.id === IModel.repositoryModelId && targetModeledElementId != sourceModel.id;
    if (isRemappedRootSubject) return;
```

I was not aware that there can only be one repository model... interesting decision with potential ramifications for forks I suppose?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2154,6 +2195,7 @@ export class IModelTransformer extends IModelExportHandler {
     }
 
     if (this._isSynchronization) {
+      // FIXME<NICK>: Should this be a nodeassert?
```
MichaelBelousov, line 2198: yes. We should stop using bentley's assert

packages/transformer/src/test/TestUtils/TimelineTestUtil.ts
```diff
@@ -251,6 +252,10 @@ export type TimelineStateChange =
           since?: number;
           initTransformer?: (transformer: IModelTransformer) => void;
           expectThrow?: boolean;
+          assertStatePostProcessChanges?: (
+            transformer: IModelTransformer
+          ) => void;
```
MichaelBelousov, line 257: what about a new subobject?
```suggestion
          assert?: {
            afterProcessChanges?(transformer: IModelTransformer): void;
          };
```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2136,9 +2162,24 @@ export class IModelTransformer extends IModelExportHandler {
       this._targetScopeProvenanceProps.version = sourceVersion;
       this._targetScopeProvenanceProps.jsonProperties.reverseSyncVersion =
         targetVersion;
+      if (this._startingChangesetIndices) {
```
MichaelBelousov, line 2165: can we not fold this into the very similar code later?

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -3458,6 +3584,261 @@ describe("IModelTransformerHub", () => {
...
+      accessToken,
+      transformerOpts: {
+        skipPropagateChangesToRootElements: true,
+      },
+    });
+
+    await tearDown();
+  });
+
+  it("should propagate changes to elements and models of rootSubject, dictionary, and realityDataSourcesElement", async () => {
```
MichaelBelousov, line 3717: is this similar enough to the previous test to consider doing:
```ts
for (const propagateRootElems of [true, false]) { 
  it(`${propagateRootElemes ? "should" : "shouldn't"} propagate changes to elements and models of rootSubject, dictionary, and realityDataSourcesElement as told", async () => {
    // use propagateRootElems to change behavior slightly
  });
}

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: not in all cases, depending on the value of the variable I mean

MichaelBelousov (approved):

Issue #155, opened 2024-02-26
Investigate having more structured testing with a case matrix


nick4598:
BranchProvenanceInitializer.test.ts already does this to an extent with sourceTargetFedGuidToAspectCountMap and the way the test cases are generated. 

A lot of test cases in IModelTransformerHub.test.ts and IModelTransformer.test.ts and may have some overlap in terms of what scenarios they test, and its hard to currently tell what all is being tested with the way they are currently setup.

A good starting point may be to take a look at the following 3 test cases:
`it("should be able to handle relationship delete using new relationship provenance method with no fedguids"`
`it("should be able to handle relationship delete using old relationship provenance method with no fedguids"`
`it("should be able to handle relationship delete using fedguids"`

and generate them programmatically.

PR #156, opened 2024-02-26
Cleanup some misc fixmes

nick4598:
I removed the fixme about structured testing and added it to work item: https://github.com/iTwin/imodel-transformer/issues/155.

Also just had some questions about some misc fixmes like transformerresumption ones.


packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -3724,9 +3721,6 @@ describe("IModelTransformerHub", () => {
     });
   }
 
-  // FIXME: As a side effect of fixing a bug in findRangeContaining, we error out with no changesummary data because we now properly skip changesetindices
```
nick4598, line 3727: Removed this because the test is no longer failing.

packages/transformer/src/IModelCloneContext.ts
```diff
@@ -313,6 +313,7 @@ export class IModelCloneContext extends IModelElementCloneContext {
 
   public override loadStateFromDb(db: SQLiteDb): void {
     super.loadStateFromDb(db);
+    // FIXME<NICK>: Can I remove this fixme, since this is related to TransformerResumption?
```
MichaelBelousov, line 316: yes

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2422,6 +2422,8 @@ export class IModelTransformer extends IModelExportHandler {
       } as SourceAndTarget);
 
     // FIXME: make importer.deleteRelationship not need full props
+    // FIXME<NICK>: Do we care that its possible an importer may want the full relationshipProps or should we just have them query for the full relationshipProps if they wanted them?
```
MichaelBelousov, line 2425: if they want it, they should query it themselves, we should avoid unnecessary information at the API level.

Unfortunately in reality, the JavaScript query API is so slow that it may less performant today. But hopefully that is addressed and eventually this change will enable a performance optimization.

packages/transformer/src/test/standalone/IModelTransformerResumption.test.ts
```diff
@@ -496,6 +496,7 @@ describe.skip("test resuming transformations", () => {
   });
 
   // FIXME: not (yet?) implemented for federation guid optimization branch
+  // FIXME<NICK>: Can I remove both the fixme and possibly even this test file?
```
MichaelBelousov, line 499: really for 1.0 you can remove the tests and all the resumption APIs

packages/transformer/src/IModelTransformer.ts
nick4598: I changed it, let me know your thoughts.

Could you explain more what you mean by "the JavaScript query API is so slow that it may less performant today". Less performant today as a result of passing all of the props? Which we're correcting in this PR now?

packages/transformer/src/test/standalone/IModelTransformerResumption.test.ts
nick4598: I took a pass at that in the commit ''delete resumption.test.ts and resumption APIs"

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: I mean that `db.query("SELECT Prop1, Prop2, Prop3 FROM bis.ElementRefersToElements WHERE ECInstanceId=?")` may be slower `than db.relationships.getInstance(id)` in all cases due to `db.query`, `db.withPreparedStatement` and `db.createQueryReader` all being slower than native today even when doing less work.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2421,14 +2421,16 @@ export class IModelTransformer extends IModelExportHandler {
         targetId: deletedRelData.targetIdInTarget,
       } as SourceAndTarget);
 
-    // FIXME: make importer.deleteRelationship not need full props
     const targetRelationship = this.targetDb.relationships.tryGetInstance(
       deletedRelData.classFullName,
       relArg
     );
```
MichaelBelousov, line 2427: can't we eliminate this query now and just pass the same args directly to `deleteRelationship`?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2494,7 +2496,10 @@ export class IModelTransformer extends IModelExportHandler {
                   ElementRefersToElements.classFullName,
                   targetRelInstanceId
                 );
-              this.importer.deleteRelationship(targetRelationship.toJSON());
+              this.importer.deleteRelationship({
+                id: targetRelationship.id,
+                classFullName: targetRelationship.classFullName,
+              });
```
MichaelBelousov, line 2502: same here, can we remove the above query now?

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: also if they are aware of that, they can just use `getInstance` themselves

packages/transformer/src/IModelTransformer.ts
nick4598: yup good call, I changed it in both places we call importer.deleteRelationship

packages/transformer/src/IModelTransformer.ts
nick4598: changed

PR #157, opened 2024-02-27
Filter the changesets as opposed to outright clearing them when the db is not at the tip.

nick4598:


packages/transformer/src/IModelTransformer.ts
```diff
@@ -2972,6 +2977,8 @@ export class IModelTransformer extends IModelExportHandler {
         (sourceElemFedGuid &&
           this._queryElemIdByFedGuid(this.targetDb, sourceElemFedGuid)) ||
         // FIXME<MIKE>: describe why it's safe to assume nothing has been deleted in provenanceDb
+        // FIXME<NICK>: Is it safe to assume nothing has been deleted in provenanceDb because the provenanceDb is (most likely?) the targetDb if we've made it to this line of code? And we're processing changes
+        // in the context of the sourceDb implying there are no changes in the targetDb to process therefore no deletes in provenanceDb?
```
MichaelBelousov, line 2981: `provenanceDb` is not necessarily the target, that's what `queryCanAccessProvenance` tells us, and it is false in a forward sync. I have to admit though, since we no longer do a real query here against the target, maybe we should reorganize this...
I see that in an earlier PR, maybe you replaced the query with `this.sourceDb.elements.getAspects` to get the `identifierValue`? I would prefer we run a more specific query rather than get the full props of all aspects and loop through them.

MichaelBelousov: maybe after we consider reorganizing, I can give the explanation for this FIXME lol?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2222,9 +2222,14 @@ export class IModelTransformer extends IModelExportHandler {
...
-      // probably wrong, and we should filter it instead
-      syncChangesetsToClear.length = 0;
+      // Only keep the changeset indices which are greater than the source, this means they haven't been processed yet.
+      syncChangesetsToClear.splice(
+        0,
+        syncChangesetsToClear.length,
+        ...syncChangesetsToClear.filter((csIndex) => {
+          return csIndex > this._startingChangesetIndices!.source;
+        })
+      );
```
MichaelBelousov, line 2232: nit: this is very confusing for a delete. Could we either stick with the `.length` idiom but calculate the length like:
```ts
syncChangesetsToClear.length = syncChangesetsToClear.findIndex(c => c > this._startingChangesetIndices.source);
```
or save the key to edit (`syncChangesetsToClearKey`) and replace the list rather than mutate it?
```ts
jsonProps[syncChangesetsToClearKey] = syncChangesetsToClear.filter();
```

packages/transformer/src/IModelTransformer.ts
nick4598: I just gave it an attempt in terms of reorganizing. Let me know if I missed what you were thinking

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2939,118 +2948,103 @@ export class IModelTransformer extends IModelExportHandler {
...
-        try {
-          element = this.sourceDb.elements.getElement(id);
-        } catch (err) {
-          return undefined;
+        if (isRelationship) {
+          try {
+            element = this.sourceDb.elements.getElement(id);
+          } catch (err) {
+            return undefined;
+          }
```
MichaelBelousov, line 2968: why not `tryGetElement`?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2939,118 +2948,103 @@ export class IModelTransformer extends IModelExportHandler {
...
-      const targetIdOfRelationshipInSource = change.TargetECInstanceId;
-      const [sourceIdInTarget, targetIdInTarget] = [
-        sourceIdOfRelationshipInSource,
-        targetIdOfRelationshipInSource,
-      ].map((id) => {
+    const changedInstanceId = change.ECInstanceId;
+    const sourceIdOfRelationshipInSource = change.SourceECInstanceId;
+    const targetIdOfRelationshipInSource = change.TargetECInstanceId;
+    const classFullName = change.$meta?.classFullName;
+    const idsInSourceDb = isRelationship
```
MichaelBelousov, line 2956: it feels like if you merge all the `isRelationship` branches into one branch it would be simpler. You would probably want to have a local function for getting the target id of an element from any source element and use it there.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2939,118 +2948,103 @@ export class IModelTransformer extends IModelExportHandler {
...
-        const fedGuid = element.federationGuid;
-        let identifierValue: string | undefined;
+        const fedGuid = isRelationship
+          ? element?.federationGuid
+          : change.FederationGuid;
         if (queryCanAccessProvenance) {
-          const aspects: ExternalSourceAspect[] =
-            this.sourceDb.elements.getAspects(
+          for await (const row of this.sourceDb.createQueryReader(
+            `SELECT esa.Identifier FROM bis.ExternalSourceAspect esa WHERE Scope.Id=:scopeId AND Kind=:kind AND Element.Id=:relatedElementId`,
```
MichaelBelousov, line 2975: add a `// TODO: clarify what happens if there are multiple (e.g. elements were merged)`
also I would go with `LIMIT 1` for now

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: much closer, have one more comment though.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2939,118 +2948,103 @@ export class IModelTransformer extends IModelExportHandler {
...
           // maybe batching these queries would perform better but we should
           // try to attach the second db and query both together anyway
-          (fedGuid && this._queryElemIdByFedGuid(this.targetDb, fedGuid))
-        );
-      });
+          (fedGuid && this._queryElemIdByFedGuid(this.targetDb, fedGuid)) ||
+          // FIXME<MIKE>: describe why it's safe to assume nothing has been deleted in provenanceDb
+          // FIXME<NICK>: Is it safe to assume nothing has been deleted in provenanceDb because the provenanceDb is (most likely?) the targetDb if we've made it to this line of code? And we're processing changes
+          // in the context of the sourceDb implying there are no changes in the targetDb to process therefore no deletes in provenanceDb?
+          (!isRelationship && this._queryProvenanceForElement(id));
```
MichaelBelousov, line 2996: so this only occurs when queryCanAccessProvenance is false (assuming `identifierValue` is never undefined in that case, which we should add an assert for), and that is false when the target is the provenanceDb (aka we're in a forward sync). We should make that more explicit imo.

This logic is very complicated so we should probably rewrite it as if/else statements with lots of intermediate variables with long descriptive names. Sorry I didn't do that in the first place.

So yes the provenanceDb is the targetDb, but I need to think about this some more :smile:, we could be forward syncing to a branch that has deleted something. I'm pretty confident that we test this case though, so probably we just need to find that test and follow it in the debugger. I'm just drawing a blank rn.

MichaelBelousov (approved):

packages/transformer/src/IModelTransformer.ts
MichaelBelousov: I was wrong (thanks @nick4598), identifierValue is from the ESAs and therefore is regularly undefined in the fedguid case.

packages/transformer/src/IModelTransformer.ts
nick4598: added local function 'getTargetIdFromSourceId'

packages/transformer/src/IModelTransformer.ts
nick4598: changed it

packages/transformer/src/IModelTransformer.ts
nick4598: I moved the queryProvenance out of the local helper function and into each respective element / relationship conditional branch. Now the local helper function is only responsible for checking by fedguid or ESA if the changedata is in the provenancedb

packages/transformer/src/IModelTransformer.ts
nick4598: changed to `tryGetElement`

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2939,118 +2948,103 @@ export class IModelTransformer extends IModelExportHandler {
     // eventually when itwin.js supports attaching a second iModelDb in JS,
     // this won't have to be a conditional part of the query, and we can always have it by attaching
     const queryCanAccessProvenance = this.sourceDb === this.provenanceDb;
```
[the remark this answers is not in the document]
nick4598: changed to `changeDataInProvenanceDb`

Issue #149, opened 2024-02-15
When calling NativeImportContext::CloneElement, the rootSubject should be special cased in the event of a remapping rule from rootSubject to some other subjectId


nick4598:
In the test "should sync Team iModels into Shared" in IModelTransformer.test.ts:

A subject is created within the shared iModel for each Team iModel. The team iModel's rootSubject is then remapped to the subject created for that team iModel with the following code: `transformerA2S.context.remapElement(IModel.rootSubjectId, subjectId)`. This is so all of the content of the team iModel, "A", finds its way under the subject "A" in the shared iModel.

Additionally, at the moment the subjectId is also added to the importer's set of doNotUpdateElementIds: `transformerA2S.importer.doNotUpdateElementIds.add(subjectId);`

Without the change to add the subjectId to the set of doNotUpdateElementIds we run into an issue where after the processAll (with team iModel as source, and shared iModel as target). The scope on the code on the subject "A" in the shared iModel changes to the id of the subject "A" itself when it should remain the same id as subject "A"s parent which is the rootSubject. 

The reason this is happening is that the rootSubject is special in that the scope of its code is the id of the rootsubject itself. This is because the rootsubject has no parent.

- ALL OTHER SUBJECTS SHOULD FOLLOW THE RULE: The scope of the code defined on me is equivalent to my parent id. 

When we remap the rootsubject: `transformerA2S.context.remapElement(IModel.rootSubjectId, subjectId)`, NativeImportContext::CloneElement mistakenly updates the scope on the code of `subjectId` to be `subjectId`. This is normal and expected behavior for all other subjects, but incorrect when the rootSubject is involved.

A change will have to be made somewhere in NativeImportContext::CloneElement.

**What does success look like?**
- Being able to comment out `transformerA2S.importer.doNotUpdateElementIds.add(subjectId);` and have the test still pass means the rootsubject was successfully special cased. 
- We should also assert that we can persist a change to some property (description for example) of the rootSubject in the team iModel "A" to the subject "A" in the shared iModel.
- We may also want to consider testing that remapping subjects other than the iModel.rootSubjectId carries parents and codes (especially their scopes) over properly. This may already be tested somewhere but I don't know.

PR #161, opened 2024-03-14
Special case root subject remapping 

jiaruiz717:
As https://github.com/iTwin/imodel-transformer/issues/149
Make root subject special cased when remapping from source to target

Fixed the bug on transformer side instead of native side as transformer is the only consumer of the corresponding native code

### DIFF
--- a/packages/transformer/src/test/standalone/IModelTransformer.test.ts
+++ b/packages/transformer/src/test/standalone/IModelTransformer.test.ts
@@ -1146,72 +1146,246 @@ describe("IModelTransformer", () => {
     iModelShared.close();
   });
 
-  it("should remapping subjects carry parents and codes over", async () => {
-    const iModelNameSource: string = "source";
-    const iModelNameTarget: string = "target";
-    const iModelFileSource: string = path.join(
+  it("remap root subject to root subject", async () => {
+    const sourceIModelName: string = "source";
+    const targetIModelName: string = "target";
+    const sourceIModelFile: string = path.join(
       outputDir,
-      `${iModelNameSource}.bim`
+      `${sourceIModelName}.bim`
     );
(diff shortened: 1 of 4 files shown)


packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -1147,22 +1147,18 @@ describe("IModelTransformer", () => {
   });
 
   it("remap root subject to root subject", async () => {
-    const sourceIModelName: string = "source";
-    const targetIModelName: string = "target";
-    const sourceIModelFile: string = path.join(
-      outputDir,
-      `${sourceIModelName}.bim`
+    const sourceIModelName: string = "source.bim";
```
nick4598, line 1150: when you're explicitly setting a variable liek you are here, no need to declare the type.

```suggestion
    const sourceIModelName = "source.bim";
```

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -1147,22 +1147,18 @@ describe("IModelTransformer", () => {
...
+    const targetIModelName: string = "target.bim";
+    const sourceIModelFile: string = path.join(outputDir, sourceIModelName);
+    const targetIModelFile: string = path.join(outputDir, targetIModelName);
+    IModelTransformerTestUtils.prepareOutputFile(
```
nick4598, line 1154: Would it work to get rid of sourceIModelName and targetIModelName?

Then just set sourceIModelFile = to the return of prepareOutputFile?

nick4598 (approved):

jiaruiz717:
Make root subject special cased when remapping from source to target

PR #163, opened 2024-03-19
Add test to reparent subjects involved in team->shared transformations

nick4598:


packages/transformer/src/IModelTransformer.ts
```diff
@@ -1526,24 +1526,24 @@ export class IModelTransformer extends IModelExportHandler {
       sourceElement,
       { binaryGeometry: this._options.cloneUsingBinaryGeometry }
     );
+
+    const targetElementId = this.context.findTargetElementId(sourceElement.id);
```
nick4598, line 1530: I also wonder, should all of this logic live in onExportElement instead?
Right now we probably only get the targetElementId if someone explicitly remapped it (I think) but its possible that rootSubject could be remapped to a nonRootSubject by virtue of having the same fedguid due to some prior explicit remapping.

PR #164, opened 2024-03-19
Check excludedElementIds before we decide to load the element in IModelExporter.exportElement

nick4598:


packages/transformer/src/IModelExporter.ts
```diff
@@ -846,6 +846,13 @@ export class IModelExporter {
       return;
     }
 
+    // Return early if the elementId is already in the excludedElementIds, that way we don't need to load the element from the db.
```
nick4598, line 849: @mindaugasdirg , per our conversation in last week's scrum you mentioned allowing an elementId to be passed to shouldExportElement I believe.

Do you see this as working for your use case? If elementIds are added to the exporter through exporter.excludeElement before processAll/processChanges is called, then the core transformer will never call `this.sourceDb.elements.getElement()` for that particular element which wasn't true previously.

mindaugasdirg: Yes, that would be perfect for our case and I see that it wouldn't require the expansion of the API.

nick4598: Great!

PR #168, opened 2024-04-15
Fix ALL lint errors and remove deprecated danglingPredecessorsBehavior

nick4598:


packages/transformer/src/EntityUnifier.ts
```diff
@@ -18,6 +18,7 @@ import {
 import {
   ConcreteEntity,
   ConcreteEntityProps,
+  // eslint-disable-next-line @typescript-eslint/no-redeclare
   Element,
```
nick4598, line 22: It was confusing to me why all the Element imports flag this, especially since it doesn't seem to be a problem in itwinjs-core.

packages/transformer/src/test/TestUtils/TimelineTestUtil.ts
```diff
@@ -36,7 +36,11 @@ import {
 import { IModelTestUtils } from "./IModelTestUtils";
 import { omit } from "@itwin/core-bentley";
 
-const { saveAndPushChanges } = IModelTestUtils;
+const saveAndPushChanges = async (
```
nick4598, line 39: Previously was getting avoid referencing unbound methods which may cause unintentional scoping of 'this'. But then if I added this: void to show that its not using this
I would get a compiler error due to binding of saveAndPushChanges to bind accessToken arg to all calls, on line 131 in hub.test.ts. So I went with this solution, resolve if fine with that.

packages/transformer/src/EntityUnifier.ts
MichaelBelousov: I think it means we're contributing the typescript lib compiler option to include the browser types

nick4598: ah interesting, well if youre fine with me individually ignoring each line feel free to resolve.

Or do we even need the browser types?

MichaelBelousov (approved):

Issue #170, opened 2024-04-22
Consider allowing transformer to run with no changesets


nick4598:
One such example where this might be useful is a filter transform. A user may wish to change the filter criteria, in which the transformer needs to be rerun to process this new filter. 

Currently services team expects to always process changesets in addition to changing the filter criteria, but it may be realistic to change the filter criteria and have no changesets to process. There are currently places where transformer exits early if there are no changesets.

One such place is if transformer._sourceDbChanges is undefined (which it currently is if there are no changesets). This would have to be changed. 

needs more thought / investigation

PR #172, opened 2024-04-25
Support 'custom' changes being added to exporter.sourceDbChanges

nick4598:


packages/transformer/src/IModelExporter.ts
```diff
@@ -1056,9 +1073,23 @@ export class ChangedInstanceIds {
   private _elementSubclassIds?: Set<string>;
   private _aspectSubclassIds?: Set<string>;
   private _relationshipSubclassIds?: Set<string>;
+  private _relationshipSubclassIdsToSkip?: Set<string>;
+  private _ecClassIdsToClassFullNames?: Map<string, string>;
+  /** c${string} is used to represent codeSpecs since they do not currently have a representation in the EntityReference class. This map holds information passed to the 'addCustom' functions. */
+  private _entityReferenceToDataMap: Map<
```
nick4598, line 1079: Wonder if I should use a BigMap?

nick4598:
One other concern I have is, does the transformation services team ever setup the exporter manually before you call the transformer? Currently you'll need to add your custom changes before you call IModelTransformer.initialize(), which I personally ran into some trouble when trying to allow myself to do that in the tests (can look at timelinetestutil to see what changes I had to make to support that)

JulijaRamoskiene:
> One other concern I have is, does the transformation services team ever setup the exporter manually before you call the transformer? Currently you'll need to add your custom changes before you call IModelTransformer.initialize(), which I personally ran into some trouble when trying to allow myself to do that in the tests (can look at timelinetestutil to see what changes I had to make to support that)

To answer your question about Exporter, yes, we do set up exporter manually.
We have our custom Exporter: https://dev.azure.com/bentleycs/iModelTechnologies/_git/imodel-transformations?path=/packages/imodel-transformations-local/src/Exporter.ts
Which is initialized and passed to transformer here: https://dev.azure.com/bentleycs/iModelTechnologies/_git/imodel-transformations?path=/packages/imodel-transformations-local/src/transformers/TransformerFactory.ts

nick4598:
> > One other concern I have is, does the transformation services team ever setup the exporter manually before you call the transformer? Currently you'll need to add your custom changes before you call IModelTransformer.initialize(), which I personally ran into some trouble when trying to allow myself to do that in the tests (can look at timelinetestutil to see what changes I had to make to support that)
> 
> To answer your question about Exporter, yes, we do set up exporter manually. We have our custom Exporter: https://dev.azure.com/bentleycs/iModelTechnologies/_git/imodel-transformations?path=/packages/imodel-transformations-local/src/Exporter.ts Which is initialized and passed to transformer here: https://dev.azure.com/bentleycs/iModelTechnologies/_git/imodel-transformations?path=/packages/imodel-transformations-local/src/transformers/TransformerFactory.ts

Thanks, I think you guys would run into the same problem I did with adding custom changes then where I had to call random transformer init functions. I changed it so that its no longer necessary and should hopefully be much easier for you guys to addcustomchanges right after you setup your exporter. 

The only real change to this is it means customchanges will now be added before any changes from changesets as opposed to customchanges being added after changes from changesets. I could see it causing some strange behavior if a custom change and a change from a changeset occurred on the same element.

packages/test-app/src/IModelHubUtils.ts
```diff
@@ -201,8 +203,8 @@ export namespace IModelHubUtils {
     return BriefcaseDb.open({
       fileName: briefcaseProps.fileName,
       readonly: briefcaseArg.briefcaseId
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
-        ? briefcaseArg.briefcaseId === BriefcaseIdValue.Unassigned
+        ? // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
```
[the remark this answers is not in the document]
nick4598: Im guessing prettier may have done this , but not sure

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2890,19 +2914,64 @@ export class IModelTransformer extends IModelExportHandler {
...
+            const classFullName =
+              this.exporter.sourceDbChanges?.getClassFullNameFromECClassId(
+                ecClassId
+              );
+            const fedGuid = customData?.federationGuid;
+            const sourceIdOfRelationshipInSource =
+              customData?.sourceIdOfRelationship;
+            const targetIdOfRelationshipInSource =
+              customData?.targetIdOfRelationship;
+            await this.processDeletedOp(
```
JulijaRamoskiene, line 2956: [Q] Why processDeleteOp() is called in processChangesets() for custom and non custom changes? Aren't deleted elements also hanlded in Exporter:exportchanges()?
image

nick4598: processDeletedOp's purpose is not to actually propagate the deletes like the exporter.exportChanges does. It mostly serves to build up the remap table for these deleted elements. We do attempt to build up the remap table in general at the top of `processChangesets`, but it wont catch any elements that no longer exist (because they were deleted).

I believe it also helps us handle the case where an element is deleted and then inserted again within a range of changesets and prevents us from accidentally deleting it from the target, when it should actually be an update in the target instead.

packages/transformer/src/IModelExporter.ts
```diff
@@ -19,6 +19,7 @@ import {
   ElementMultiAspect,
   ElementRefersToElements,
```
JulijaRamoskiene, line 20: **Requirements from my side:**  
- Possibility to add custom changes (SourceDbChanges). 
- Custom changes should not be overridden by changes in changesets. 
- It is preferable to add changes after those are initialized from changesets. This will provide possibility to add custom changes based on changes in changesets.
- It should be possible to run processChanges transformation with custom changes when there are no changes in source. This is least priority (although still important) and will be tackled in separate PR:  https://github.com/iTwin/imodel-transformer/issues/170

packages/transformer/src/IModelExporter.ts
```diff
@@ -1145,6 +1205,132 @@ export class ChangedInstanceIds {
...
+   * from the set of updatedIds and add it to the set of deletedIds for the appropriate class type.
+   * @note This method will not accept changes to relationship classes. Use 'addCustomRelationshipChange' instead.
+   * @note In most cases, this method does not need to be called. Its only for consumers to mimic changes as if they were found in a changeset, which should only be useful in certain cases such as the changing of filter criteria for a preexisting master branch relationship.
+   * @throws if the ecClassId is a relationship classId
+   * @param ecClassId class id of the custom change
+   * @param changeType insert, update or delete
+   * @param id ECInstanceID of the custom change
+   * @param federationGuid federationGuid defined on the element
+   */
+  public async addCustomChange(
```
JulijaRamoskiene, line 1220: - It would be great to set custom changes without passing ecClassId (and federation guid if possible).
- there is no indication if custom change is not added (e.g. class is not mapped). Suggest to throw error instead of silently ignoring invalid custom change. 
- Currently it is not registering custom changes for repository model, can be fixed by : 
```
 addECClassIdsToSet(this._modelSubclassIds, "BisCore.RepositoryModel"),
```

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -806,6 +806,7 @@ describe("IModelTransformerHub", () => {
       assert.equal(7, count(sourceDb, PhysicalModel.classFullName));
       // 60 elements added
       assert.equal(185, count(sourceDb, PhysicalObject.classFullName));
+      // transformer.exporter["_initialized"] = false;
```
JulijaRamoskiene, line 809: this commented line can be removed

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -1804,6 +1813,213 @@ describe("IModelTransformerHub", () => {
     }
   });
 
+  it.only("should propagate custom inserts and custom deletes", async () => {
```
JulijaRamoskiene, line 1816: More tests will be needed e.g. to add / delete model

JulijaRamoskiene: I feel like it is easier to add tests on my side, as I have "filtered" target. Here it is more  complicated, eg: to test insert you need to delete that element from target first. Is there a way to simplify this?

JulijaRamoskiene (changes requested):

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
nick4598: done

packages/transformer/src/IModelExporter.ts
nick4598: Regarding this requirement: "Custom changes should not be overridden by changes in changesets."

Is this still a requirement if you will be adding custom changes after changes in changesets are processed and you're using the sourceDbChanges as guidance for what to add? 

I think its possible to do custom changes both before and after at the moment, so maybe we should either look to make it only after or make it so that ones are added before are the single source of truth? I suppose I lean towards only after changesets are processed if thats possible, haven't looked into it yet.

packages/transformer/src/IModelExporter.ts
nick4598: Regarding the necessity of classId. It doesn't seem necessary, I am fine with trusting callers of the API to call the appropriate function. I converted addCustomChange to take in an EntityType instead of the classId and federationGuid. 
 
I do still wonder about relationships though and deleting them. Do you guys know whether or not you expect to have to add custom relationship changes? onDeleteRelationship seems to use the classFullName
 
Either way maybe we could just keep addCustomRelationshipChange as a slightly different API in that it would require classId if we had to...

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2890,19 +2914,67 @@ export class IModelTransformer extends IModelExportHandler {
...
 
       csReader.close();
     }
+
+    // This loop is to process all custom deleteIds. Unclear if the special logic is still necessary for relationships or not (TODO!!). For all other entities, we assume that the element is still present in the sourceDb because it is not
+    // a real delete and instead a simulated delete to update filtering criteria between source and target. Since the element is still present, we do not need to call processDeletedOp to find the corresponding targetId.
+    // We can instead rely on `forEachTrackedElement` at the top of processChangesets to find the corresponding targetId.
+    // Note this also assumes we don't need to handle entity recreation for these custom deletes. I.e. a caller of API would not be able to add a custom delete for an entity that was recreated.
+    // a delete followed by an insert.
+    // ASSUME: If a changeset has a deleteId then custom change will never reference it. Is this still true if it was re-inserted? (TODO!!)
```
nick4598, line 2938: Im still unsure about entity recreation.

Basically a delete followed by an insert of the same entity (They would have different IDs, but may be considered the same if they had the same federationGuid).

What would you guys do with that with regards to customChanges?

Would it be possible for us to have that in our changesets (entity recreation) and then also have you decide that the insert (after the delete) should not make it to the target?

I think in theory the target would have the entity in it already... so you could potentially have a scenario where the entity was recreated in the source, but you know you also want to simulate a delete so that we get rid of it in the target. That scenario seems difficult to me.

(im not sure if im making sense)

Perhaps one option is to not let customChanges coexist with actual changes? Not sure how annoying that would be for you because I could see it meaning you have to run two transforms, one to get the changesets processed adn then one to update filter criteria. 

Theres probably another option we could come up with, but a little unsure how you guys plan to use customChanges in this specific scenario.

JulijaRamoskiene: **"Would it be possible for us to have that in our changesets (entity recreation) and then also have you decide that the insert (after the delete) should not make it to the target?"**
Yes, this is very possible. I think we should also add test case for such scenario. When filtering parameters change, it is possible that all elements that were previously added to target, now need to be removed. If e.g. in 1 transformation elements of "Category1" were exported, and now filtering parameters changed to export only "category 2" elements. 


**"Perhaps one option is to not let customChanges coexist with actual changes? Not sure how annoying that would be for you because I could see it meaning you have to run two transforms, one to get the changesets processed and then one to update filter criteria."** 

Not sure I fully understand how it would work, we would need to discuss this possibility a bit more.

Considerations I have about running 2 separate transformations, are:
- there is no support to run 2 transformation at the moment
- I don't plan to store/use old filter criteria, so first transformation would run with new new filter criteria. and then second transformation would do additional inserts / deletes as custom changes.  Would need to make more testing to see how that would work.   
- Support to run transformation when there are no changes in source would be needed for second transformation (as there would be only custom changes). But this will be needed anyway, so probably not a big blocker.

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
nick4598: I suppose instead of deleting it from target, I could exclude the things I wnat to later add as custom changes using excludeElement? Would have to expose that in the timelinetestutil.ts I think though. I could see that making the test a bit 'simpler' but idk by how much.

When you say you have the 'filtered' target. Do you mean you are pushing to the repo some small test iModel? Maybe even two? the source and the filtered target?

I'm not opposed to having something like that as I know other repos do it like itwinjs-core. If we were to do it though im unsure about where to draw the line. It'd probably be overkill to do it for every test of course but maybe here and there seems reasonable

packages/transformer/src/IModelExporter.ts
```diff
@@ -580,7 +596,7 @@ export class IModelExporter {
   public async exportCodeSpecByName(codeSpecName: string): Promise<void> {
     const codeSpec: CodeSpec = this.sourceDb.codeSpecs.getByName(codeSpecName);
     let isUpdate: boolean | undefined;
-    if (undefined !== this._sourceDbChanges) {
+    if (this._sourceDbChanges !== undefined && !this._sourceDbChanges.isEmpty) {
```
JulijaRamoskiene, line 599: [Q] Can you explain more about this change: does this change means that all codespecs will be exported when sourceDbChanges are empty? Why we want that?
Is it possible to instead add needed code Specs to sourceDbChanges,  when custom changes are added? Same as e.g. parent models would be marked as updated when new model is inserted.

JulijaRamoskiene: Same question for other places like exportModelContainer()

packages/transformer/src/IModelExporter.ts
```diff
@@ -1145,6 +1228,117 @@ export class ChangedInstanceIds {
...
+        `Misuse. id: ${id} is a relationship. Use 'addCustomRelationshipChange' instead.`
+      );
+    }
+    this._hasCustomChanges = true;
+
+    switch (entityType) {
+      case "codeSpec":
+        this.handleChange(this.codeSpec, changeType, id);
+        break;
+      case "model":
```
JulijaRamoskiene, line 1260: need to update parent model (models? ) as updated when child model is marked as inserted.

packages/transformer/src/IModelExporter.ts
```diff
@@ -1145,6 +1228,117 @@ export class ChangedInstanceIds {
...
+        this.handleChange(this.model, changeType, id);
+        break;
+      case "element": // When element we need to mark the element's model as updated otherwise this fails.
+        this.addModelToUpdated(id);
```
JulijaRamoskiene, line 1264: [Q] I was not aware that element model needs to be marked as updated. Is it enough to mark element model as updated, or all parent models need to be marked as well?

packages/transformer/src/IModelExporter.ts
JulijaRamoskiene: custom relationship changes would be added only in case if it is required when elements / models are marked as inserted/deleted. Is it needed in such cases or relationships will be handled automatically?

packages/transformer/src/IModelExporter.ts
```diff
@@ -1145,6 +1228,117 @@ export class ChangedInstanceIds {
...
+  ): Promise<void> {
+    if (!this._ecClassIdsInitialized) await this.setupECClassIds();
+    if (entityType === "relationship") {
+      throw new Error(
+        `Misuse. id: ${id} is a relationship. Use 'addCustomRelationshipChange' instead.`
+      );
+    }
+    this._hasCustomChanges = true;
+
+    switch (entityType) {
```
JulijaRamoskiene, line 1256: Suggest to have multiple small methods instead of one generic one `addCustomChange`, e.g.:
- `addCustomElementChange` or `addElementChange()`
- `addCustomModelChange()` / `addModelChange()`
- `addCustomCodeSpecChange()`
- `addCustomAspectChange()`

packages/transformer/src/IModelExporter.ts
```diff
@@ -1145,6 +1228,117 @@ export class ChangedInstanceIds {
...
+   * @param changeType insert, update or delete
+   * @param id ECInstanceID of the custom change
+   * @param federationGuid federationGuid defined on the element
+   */
+  public async addCustomChange(
+    entityType: ChangedInstanceType,
+    changeType: SqliteChangeOp,
+    id: Id64String // TODO: SUPPORT BULK ADDS (per entity type okay?)
+  ): Promise<void> {
+    if (!this._ecClassIdsInitialized) await this.setupECClassIds();
```
JulijaRamoskiene, line 1248: is it needed to setup ECClass ids here? It is also called in addChange()

packages/transformer/src/IModelExporter.ts
```diff
@@ -1145,6 +1228,117 @@ export class ChangedInstanceIds {
...
+      {
+        sourceIdOfRelationship: sourceECInstanceId,
+        targetIdOfRelationship: targetECInstanceId,
+        ecClassId,
+      }
+    );
+    this.handleChange(this.relationship, changeType, id);
+  }
+
+  public getClassFullNameFromECClassId(
```
JulijaRamoskiene, line 1336: I think `getClassFullNameFromECClassId()` should not be implemented on  `ChangedInstanceIds` as public method. It could be some private method / variable, but it looks strange being public. Also, only populated classes can be resolved, not sure if it might introduce errors.

JulijaRamoskiene: Sample of class that would not be resolved: "BisCore::RepositoryModel"

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2890,19 +2959,35 @@ export class IModelTransformer extends IModelExportHandler {
         )
           continue;
         await this.processDeletedOp(
-          change,
+          change.ECInstanceId,
+          change.$meta?.classFullName ?? "",
```
JulijaRamoskiene, line 2963: - What is means when empty string is passed for classFullName? Will it still produce expected results?   
- It could be a good idea to split `processDeletedOp()` into `processRelationshipDeleteOp()` and `processElementDeleteOpt()`, as it seems that separate sets of variables are needed to process these 2, e.g. className, sourceIdOfRelationshipInSource and targetIdOfRelationshipInSource  are not needed when processing element delete operation.

packages/transformer/src/IModelExporter.ts
nick4598: Currently relationships will not be handled automatically and will need to be explicitly added to the sourceDbChanges, atleast inserts and updates it seems. Not sure about deletes (whether element deletes delete relationships automatically)

Im open to looking for relationships in addCustomElementChanges and adding those ids, but I do wonder if there would ever be a case where you just need to insert a relationship (relationship was filtered out/excluded?) 
 
or you wouldn't want relationship to come over just because you inserted the element.

packages/transformer/src/IModelExporter.ts
nick4598: Originally I was doing this because I was always defining sourceDbChanges, but you're right that this is a logical error. I didnt understand why tests weren't failing ( because we have some that assert no changes) until I realized there are multiple places we can stop ourselves from exporting changes and I was getting saved by those future checks. Created work item to improve that test

Ive stopped using isEmpty except for in the transformer and sourceDbChanges can now be undefined if exporter is initialized with {}. 

```
 private async _tryInitChangesetData(args?: ProcessChangesOptions) {
    if (
      !args ||
      this.sourceDb.iTwinId === undefined ||
      this.sourceDb.changeset.index === undefined
    ) {
      this._sourceChangeDataState = "unconnected";
      return;
    }

    const noChanges =
      this.synchronizationVersion.index === this.sourceDb.changeset.index;
    if (noChanges) {
      this._sourceChangeDataState = "no-changes";
      this._csFileProps = [];
      return;
    }
```

I think probably in the case where no changesets and args are passed i.e. argsForProcessChanges, we will just set csFileProps to [] and sourceDbChanges will be defined, because we'll pass [] to exporter.initialize in transformer.

packages/transformer/src/IModelExporter.ts
```diff
@@ -355,19 +360,30 @@ export class IModelExporter {
...
 
+  /**
+   * This function is called by the transformer as it is about to process the changesets passed to it in [[IModelTransformOptions.argsForProcessChanges]].
+   * This would be after the exporter has already processed the same set of changesets passed to the transformer in [[IModelTransformOptions.argsForProcessChanges]].
+   * This function should be used to modify the exporter's sourceDbChanges, if necessary, using [[ChangedInstanceIds.addCustomChange]]. See [[ChangedInstanceIds.addCustomChange]] for more information.
+   * @note It is guaranteed that exporter.sourceDbChanges will be defined and populated with the changesets passed to the transformer, if any, when this function is called by the transformer.
+   * @note The transformer will have built up the remap table between the source and target iModels before calling this function. This means that functions like [[IModelTransformer.context.findTargetElementId]] will return meaningful results.
+   * @note Its expected that this function be overridden by a subclass of exporter if it needs to modify sourceDbChanges.
+   */
+  public addCustomChangesCallback(): void {}
```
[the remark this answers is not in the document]
nick4598: changed to addCustomChanges

packages/transformer/src/IModelExporter.ts
nick4598: It seems to be enough that only the element's model is marked as updated. I don't see any logic which prevents us from calling exportSubModels if the parentmodel has no changes.

packages/transformer/src/IModelTransformer.ts
nick4598: I explored your first bullet point and seems like the answer is no. We would fail if we called 'queryProvenanceForRelationship' with "" and we would also fail when wewent to query the deletedSourceRelationshipData in order to delete it in the target. For now I've decided to log an error if this scenario occurs warnning that the relationship delete may not propagate. Im not sure its worth throwing the entire transformer run away for, but maybe it is. Right now I've decided to not do that.

nick4598: Refactored processDeletedOp as well

packages/transformer/src/IModelExporter.ts
nick4598: switched to private, added classFullName to the customdatamap for relationships

packages/transformer/src/IModelExporter.ts
nick4598: with the change to ahve a individual function for element, model, etc. No need to setup ecclassIds anymore

packages/transformer/src/IModelExporter.ts
nick4598: added small methods for each entity type

packages/transformer/src/IModelExporter.ts
JulijaRamoskiene: It starts model exports from 'IModel.repositoryModelId' here:

image

Child model export is gated by parent models being updated in 'exportModelContents':
image


To insert model, current implementation requires that:
- model id is added to insertedIds
- matching partition element id is added to inserted ids
- parent models are marked as updated

As you mentioned before, with new custom changes API it can be achieved with 2 calls:   
```typescript
exporter.sourceDbChanges?.addCustomModelChange("Inserted", modelIdtoInsert);

// will add partition element to inserted ids and will add parent model (model of partition element) to updated ids  

exporter.sourceDbChanges?.addCustomElementChange("Inserted", modelIdtoInsert);

```
 So it seems to work fine with current implementation (parent model is marked as updated), only concerns I have:
- Is it enough to mark partition element model as updated, or whole hierarchy (all parents) need to be marked as updated? 
- I agree with you that "inserting a model on its own with no associated element" is not useful, at least in context of transformer, as only adding custom change to insert model without inserting partition element will always throw error. Should addCustomModelChange() implementation call addCustomElementChange() for partition element? 
image

nick4598: >Is it enough to mark partition element model as updated, or whole hierarchy (all parents) need to be marked as updated?

I think since exportSubModels is called for every model, whole hierarchy would not be necessary. Any specific concerns you have there?
ALl it takes to call exportSubModels is for the modeledElement to pass the `shouldExportElement` check, which doesn't trake a look at any sourceDbCHanges. Its mostly concerned with any exclusions from `exporter.excludeElement` or any element classes excluded. I think failing the shouldExportElement check resulting in not calling exportSubModels makes sense.

>I agree with you that "inserting a model on its own with no associated element" is not useful, at least in context of transformer, as only adding custom change to insert model without inserting partition element will always throw error. Should addCustomModelChange() implementation call addCustomElementChange() for partition element?

Im in favor of adding more logic to addCustomModelChange. One thing that concerns me is, would you as the user of the API know that an element you want to add is actually the modeledelement of some model? I guess in the worst case you would probably addCustomModelChange and addCustomElementChange both with the same id and same changetype (both inserted or both deleted) and nothing out of the ordinarY would happen. If user happens to do one inserted and one deleted im fine considering that undefined behavior.

Also side thought: Does it ever make sense that you need to specifically add a custom change of type "Updated"? To me it seems like Inserted or Deleted would be all you need. Wondering if its worth narrowing the changetype for the addcustomchanges functions for that reason. I guess I could see a case where element A makes it through first transform, element A stops making it through but you dont delete it from the target. Then element A suddenly makes it through again but you need to get the new changes to A, if any, over to the target. That would be a custom updated case (not sure if real).

packages/transformer/src/IModelExporter.ts
```diff
@@ -1027,6 +1042,33 @@ export class ChangedInstanceOps {
...
+  public get isEmpty(): boolean {
+    return (
+      0 === this.insertIds.size &&
+      0 === this.updateIds.size &&
+      0 === this.deleteIds.size
+    );
+  }
+}
+
+export type ChangedInstanceType =
```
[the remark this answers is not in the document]
nick4598: Removed it

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2789,8 +2839,14 @@ export class IModelTransformer extends IModelExportHandler {
         this.context.remapElement(sourceElementId, targetElementId);
       }
     );
-    if (this._csFileProps === undefined || this._csFileProps.length === 0)
-      return;
+    this.exporter.addCustomChanges();
+
+    if (this._csFileProps === undefined || this._csFileProps.length === 0) {
+      if (this.exporter.sourceDbChanges?.isEmpty) return;
```
JulijaRamoskiene, line 2845: `this._sourceChangeDataState` might end up being set to  "has-changes" when `this.exporter.sourceDbChanges` is `undefined` due to this if: `if (this.exporter.sourceDbChanges?.isEmpty)`.

common/api/imodel-transformer.api.md
```diff
@@ -49,8 +73,14 @@ export class ChangedInstanceIds {
     element: ChangedInstanceOps;
     // (undocumented)
     font: ChangedInstanceOps;
+    // @beta
+    getCustomRelationshipDataFromId(id: Id64String): ChangedInstanceCustomRelationshipData | undefined;
+    // (undocumented)
+    get hasCustomChanges(): boolean;
```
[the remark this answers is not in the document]
nick4598: Changed to hasCustomRelationshipChanges for now, its currently only useful for relationship deletes (maybe.. still unsure about those)

packages/transformer/src/IModelTransformer.ts
nick4598: blah good catch, now returning if undefined || empty

packages/transformer/src/IModelExporter.ts
```diff
@@ -454,7 +479,7 @@ export class IModelExporter {
 
     await this.exportCodeSpecs();
     await this.exportFonts();
-    if (initOpts.skipPropagateChangesToRootElements) {
+    if (initOpts?.skipPropagateChangesToRootElements) {
```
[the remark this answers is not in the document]
nick4598: removed

packages/transformer/src/IModelExporter.ts
```diff
@@ -368,6 +374,17 @@ export class IModelExporter {
...
+   * This would be after the exporter has already processed the same set of changesets passed to the transformer in [[IModelTransformOptions.argsForProcessChanges]].
+   * This function should be used to modify the exporter's sourceDbChanges, if necessary, using [[ChangedInstanceIds.addCustomChange]]. See [[ChangedInstanceIds.addCustomChange]] for more information.
+   * @note [[IModelExporter.sourceDbChanges]] will only be defined if the transformer was called with [[IModelTransformOptions.argsForProcessChanges]].
+   * @note If defined, sourceDbChanges will already be populated with the changesets passed to the transformer, if any when this function is called by the transformer.
+   * @note The transformer will have built up the remap table between the source and target iModels before calling this function. This means that functions like [[IModelTransformer.context.findTargetElementId]] will return meaningful results.
+   * @note Its expected that this function be overridden by a subclass of exporter if it needs to modify sourceDbChanges.
+   */
+  public addCustomChanges(): void {}
```
JulijaRamoskiene, line 386: The more I look at my implementation, the more I see that there would be less code on my side if 'addCustomChanges' was moved to `IModelTransformer` class.
Suggest these changes to `Exporter.addCustomChanges()`
 - move this function to `IModelTransformer`
 - make it `async`
 - change access modifier to `protected`, this will enforce that it is meant to be overridden or called only in `IModelTransformer` and its subclasses.
 - pass 'sourceDbChanges' as function parameter, it will be more intuitive for devs to understand, where changes should be added.
```
IModelTransformer {
    protected async addCustomChanges(sourceDbChanges: ChangedInstanceIds): Promise<void> {}
}
```

More reasons to justify these changes: 
- Transformer is the one that needs to control when custom changes are added. he should be the only one that can call `addCustomChanges()` which will be enforced if it becomes protected.
- Users that want to add custom changes will be able to do it without implementing custom exporter
- In my implementation I am using `IModelTransformer.context` property to check if element already exist in target, this property is available in `IModelTransformer` but not in Exporter.

packages/transformer/src/IModelExporter.ts
```diff
@@ -1027,6 +1052,26 @@ export class ChangedInstanceOps {
         val.delete.forEach((id: Id64String) => this.deleteIds.add(id));
     }
   }
+
+  public get isEmpty(): boolean {
```
JulijaRamoskiene, line 1056: Tried to use this from exporter and have suggestion: since `Exporter.sourceDbChanges` property is optional, I think `isEmpty` should be refactored into `hasChanges`, then instead of this
```
const isEmpty = this.exporter.sourceDbChanges === undefined || this.exporter.sourceDbChanges.isEmpty;
const hasChanges = this.exporter.sourceDbChanges !== undefined && !this.exporter.sourceDbChanges.isEmpty
```
we could do this:
```
const isEmpty = !this.exporter.sourceDbChanges?.hasChanges;
const hasChanges = this.exporter.sourceDbChanges?.hasChanges
```

Another solution could be to make `Exporter.sourceDbChanges` required, but this seems like a lot more work.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2789,8 +2839,18 @@ export class IModelTransformer extends IModelExportHandler {
         this.context.remapElement(sourceElementId, targetElementId);
       }
     );
-    if (this._csFileProps === undefined || this._csFileProps.length === 0)
-      return;
+    this.exporter.addCustomChanges();
```
JulijaRamoskiene, line 2842: In my implementation I need to throw error if there are no changes (in source db&custom), do you have suggestions for best way I could achieve it? My first attempt looks something like this:
```
      await this.initialize();
      const isEmpty = this.exporter.sourceDbChanges === undefined || this.exporter.sourceDbChanges.isEmpty;
      if (isEmpty)
          throw xxxx

```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2889,145 +2950,212 @@ export class IModelTransformer extends IModelExportHandler {
...
+    changedInstanceId: Id64String,
+    classFullName: string,
+    sourceIdOfRelationshipInSource: Id64String,
+    targetIdOfRelationshipInSource: Id64String
+  ) {
+    // we need a connected iModel with changes to remap elements with deletions
+    const notConnectedModel = this.sourceDb.iTwinId === undefined;
+    const noChanges =
+      this.synchronizationVersion.index === this.sourceDb.changeset.index &&
+      this.exporter.sourceDbChanges?.isEmpty;
```
JulijaRamoskiene, line 3023:
```
 const noChanges =
      this.synchronizationVersion.index === this.sourceDb.changeset.index &&
      this.exporter.sourceDbChanges?.isEmpty;
```
`noChanges` will be always `false` when this.exporter.sourceDbChanges is `undefined`. 

Suuggest this: https://github.com/iTwin/imodel-transformer/pull/172/files#r1910400545

nick4598: Gah I am getting tripped up by so many of these conditional operators.

I think your suggestion is correct but how do you feel about the more explicit version:
```
    const noChanges =
      this.synchronizationVersion.index === this.sourceDb.changeset.index &&
      (this.exporter.sourceDbChanges === undefined || !this.exporter.sourceDbChanges.hasChanges);
```

Instead of `!this.exporter.sourceDbChanges?.hasChanges`

packages/transformer/src/IModelExporter.ts
nick4598: Moved to transformer class

packages/transformer/src/IModelExporter.ts
nick4598: changed to hasChanges

packages/transformer/src/IModelTransformer.ts
nick4598: That seems appropriate, would be slightly different with the change from isEmpty to hasChanges

packages/transformer/src/IModelTransformer.ts
JulijaRamoskiene: Seems ok.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2889,145 +2951,227 @@ export class IModelTransformer extends IModelExportHandler {
...
+   * This would be after the exporter has already processed the same set of changesets passed to the transformer in [[IModelTransformOptions.argsForProcessChanges]].
+   * This function should be used to modify the exporter's sourceDbChanges, if necessary, using [[ChangedInstanceIds.addCustomChange]]. See [[ChangedInstanceIds.addCustomChange]] for more information.
+   * @param sourceDbChanges will only be defined if the transformer was called with [[IModelTransformOptions.argsForProcessChanges]].
+   * @note If defined, sourceDbChanges will already be populated with the changesets passed to the transformer, if any when this function is called by the transformer.
+   * @note The transformer will have built up the remap table between the source and target iModels before calling this function. This means that functions like [[IModelTransformer.context.findTargetElementId]] will return meaningful results.
+   * @note Its expected that this function be overridden by a subclass of transformer if it needs to modify sourceDbChanges.
+   */
+  protected async addCustomChanges(
+    _sourceDbChanges?: ChangedInstanceIds
```
JulijaRamoskiene, line 3070: `_sourceDbChanges?` should be required. As API user I have no idea how to handle undefined source db changes. It is impossible to add custom changes and I am too worried to throw error (what if it is expected). 
It should be transformers responsibility to handle case when source db change is undefined and act accordingly (throw / not to call addCustomChanges / initalize custom changes / something else).

packages/transformer/src/IModelExporter.ts
```diff
@@ -1145,6 +1220,142 @@ export class ChangedInstanceIds {
...
+  /**
+   * Adds the provided change to the model changes maintained by this instance of ChangedInstanceIds.
+   * Also adds the model's modeledElement to the element changes. This is to ensure the changes from the model and its modeledElement get exported together.
+   * If the same ECInstanceId is seen multiple times, the changedInstanceIds will be modified accordingly, i.e. if an id 'x' was updated but now we see 'x' was deleted, we will remove 'x'
+   * from the set of updatedIds and add it to the set of deletedIds for the appropriate class type.
+   * @note It is the responsibility of the caller to ensure that the provided id is, in fact a model.
+   * @note In most cases, this method does not need to be called. Its only for consumers to mimic changes as if they were found in a changeset, which should only be useful in certain cases such as the changing of filter criteria for a preexisting master branch relationship.
+   * @beta
+   */
+  public addCustomModelChange(changeType: SqliteChangeOp, ids: Id64Arg): void {
```
JulijaRamoskiene, line 1269: Continuing this discussion: https://github.com/iTwin/imodel-transformer/pull/172/files/e576fa6b45d47d3996366ff9a910a9d5369ba892#r1888402014

I realized where was my logical mistake was in previous discussion, these are fixed requirements of how model can be inserted via custom changes API:

- model id is added to insertedIds
- modeled element id is added to inserted ids (it is same as model id). This is needed as model can not be created without modeled element.
- model of modeled element is added to updated / inserted ids. This is needed so that modeled element is exported to target in exportModeledContents()


Turns out `model of modeled element` needs to be marked as changed (not `parent model of inserted model` as I thought previously). I got mixed up as those are the same in my test data (both equal to RepositoryModel).

Suggest to update current implementation to call `addCustomElementChange` instead of 'handleChange' with modelet element id:
```typescript
    addCustomModelChange(changeType, ids) {
        for (const id of core_bentley_1.Id64.iterable(ids)) {
            this.handleChange(this.model, changeType, id);
            // Also add the model's modeledElement to the element changes. The modeledElement and model go hand in hand.
            addCustomElementChange(changeType, id);
        }
    }
```

This is test that I used for my investigation, which is failing with current implementation: 
```typescript
  it.only("should insert Model", async () => {
    const masterIModelName = "Master";
    const masterSeedFileName = path.join(outputDir, `${masterIModelName}.bim`);
    if (IModelJsFs.existsSync(masterSeedFileName))
      IModelJsFs.removeSync(masterSeedFileName);
    const masterSeedState = { 1: 1, 2: 1, 20: 1, 21: 1, 40: 1, 41: 2, 42: 3 };
    const masterSeedDb = SnapshotDb.createEmpty(masterSeedFileName, {
      rootSubject: { name: masterIModelName },
    });
    // eslint-disable-next-line deprecation/deprecation
    masterSeedDb.nativeDb.setITwinId(iTwinId); // workaround for "ContextId was not properly setup in the checkpoint" issue
    const schemaPathForMultiAspect =
      IModelTransformerTestUtils.getPathToSchemaWithMultiAspect();
    await masterSeedDb.importSchemas([schemaPathForMultiAspect]);

    populateTimelineSeed(masterSeedDb, masterSeedState);

    const masterSeed: TimelineIModelState = {
      // HACK: we know this will only be used for seeding via its path and performCheckpoint
      db: masterSeedDb as any as BriefcaseDb,
      id: "master-seed",
      state: masterSeedState,
    };

    let physicalModelIdInSource: Id64String | undefined;
    let modelUnderRepositoryModel: Id64String | undefined;
    const timeline: Timeline = [
      { master: { seed: masterSeed } }, // masterSeedState is above
      { branch1: { branch: "master" } },
      { master: { 100: 100, 101: 101 } },
      {
        master: {
          manualUpdate(db) {
            physicalModelIdInSource = PhysicalModel.insert(
              db,
              IModel.rootSubjectId,
              "MyPhysicalModel"
            );
            modelUnderRepositoryModel = DefinitionModel.insert(
              db,
              IModel.rootSubjectId,
              "MyModelUnderRepositoryModel"
            );
          },
        },
      },
      { branch1: { sync: ["master"] } }, // master->branch1 forward sync to pick up relationship change
      {
        branch1: {
          // delete model and element from branch so that we can attempt to add it back in as a custom 'Inserted' change
          manualUpdate(db) {
            const physicalPartitionIdInTarget =
              IModelTestUtils.queryByCodeValue(db, "MyPhysicalModel");
            expect(physicalPartitionIdInTarget).to.not.equal(Id64.invalid);
            db.models.deleteModel(physicalPartitionIdInTarget);
            db.elements.deleteElement(physicalPartitionIdInTarget);
            const modelUnderRepositoryModelId =
              IModelTestUtils.queryByCodeValue(
                db,
                "MyModelUnderRepositoryModel"
              );
            expect(modelUnderRepositoryModelId).to.not.equal(Id64.invalid);
            db.models.deleteModel(modelUnderRepositoryModelId);
            db.elements.deleteElement(modelUnderRepositoryModelId);
          },
        },
      },
      {
        assert({ branch1 }) {
          // Extra assert to make sure relationship and element are deleted in branch1
         
          // assume if element is gone, aspect is gone
          const physicalPartitionIdInTarget = IModelTestUtils.queryByCodeValue(
            branch1.db,
            "MyPhysicalModel"
          );
          expect(physicalPartitionIdInTarget).to.equal(Id64.invalid);
          const modelUnderRepositoryModelInTarget =
            IModelTestUtils.queryByCodeValue(
              branch1.db,
              "MyModelUnderRepositoryModel"
            );
          expect(modelUnderRepositoryModelInTarget).to.equal(Id64.invalid);
        },
      },
      {
        branch1: {
          sync: [
            "master",
            {
              init: {
                afterInitializeExporter: async (exporter) => {
                  exporter.sourceDbChanges?.addCustomModelChange(
                    "Inserted",
                    physicalModelIdInSource!
                  );
                  exporter.sourceDbChanges?.addCustomModelChange(
                    "Inserted",
                    modelUnderRepositoryModel!
                  );
                },
              },
            },
          ],
        },
      },
      {
        assert({ branch1 }) {
          const physicalPartitionIdInTarget = IModelTestUtils.queryByCodeValue(
            branch1.db,
            "MyPhysicalModel"
          );
          expect(physicalPartitionIdInTarget).to.not.equal(Id64.invalid);
          expect(branch1.db.elements.getElement(physicalPartitionIdInTarget)).to
            .not.be.undefined;
          expect(branch1.db.models.getModel(physicalPartitionIdInTarget)).to.not
            .be.undefined;
          const modelUnderRepositoryModelInTarget =
            IModelTestUtils.queryByCodeValue(
              branch1.db,
              "MyModelUnderRepositoryModel"
            );
          expect(modelUnderRepositoryModelInTarget).to.not.equal(Id64.invalid);
          expect(
            branch1.db.elements.getElement(modelUnderRepositoryModelInTarget)
          ).to.not.be.undefined;
          expect(branch1.db.models.getModel(modelUnderRepositoryModelInTarget))
            .to.not.be.undefined;
        },
      }
    ];
    const { tearDown } = await runTimeline(timeline, {
      iTwinId,
      accessToken,
    });
    await tearDown();
  });
```

packages/transformer/src/IModelExporter.ts
JulijaRamoskiene: Continuing this discussion here: https://github.com/iTwin/imodel-transformer/pull/172/files#r1920173293

packages/transformer/src/IModelExporter.ts
```diff
@@ -1145,6 +1220,142 @@ export class ChangedInstanceIds {
...
+   * @note In most cases, this method does not need to be called. Its only for consumers to mimic changes as if they were found in a changeset, which should only be useful in certain cases such as the changing of filter criteria for a preexisting master branch relationship.
+   * @beta
+   */
+  public addCustomElementChange(
+    changeType: SqliteChangeOp,
+    ids: Id64Arg
+  ): void {
+    // if delete unnecessary?
+    for (const id of Id64.iterable(ids)) {
+      this.addModelToUpdated(id);
```
JulijaRamoskiene, line 1238: To improve performance, consider querying models of all elements at once in `addModelToUpdated()`

packages/transformer/src/IModelTransformer.ts
nick4598: made sourceDbChanges required, and only call addCustomChanges if its defined.

packages/transformer/src/IModelExporter.ts
nick4598: So does this mean the 'modelUnderRepositoryModel' is where we fail? `expect(modelUnderRepositoryModelInTarget).to.not.equal(Id64.invalid);`? because its modeled element has a model which is the repositoryModel, and we didn't add it to the updated?

I didn't realize the modeled element could have a model that wasn't the model it was modelling... I started looking at the documentation for model:

Im guessing 'modeled element's model' is being refereenced by this line here? "The Element modeling the car-as-a-whole is also in a Model. What Element is that Model sub-modeling? BIS escapes from infinite regression by defining a special RepositoryModel that is not required to sub-model some other Element." [docs](https://www.itwinjs.org/bis/guide/intro/modeling-with-bis/#model)

Is that right?

And to take it further the modeled element's model would be found right here on an element:
image
[image taken from here](https://www.itwinjs.org/reference/core-backend/elements/element/)

Is that also right?

packages/transformer/src/IModelExporter.ts
nick4598: Switched to one query for all the elements

packages/transformer/src/IModelExporter.ts
nick4598: Thanks for figuring this out. Ive made the change that you suggested and also changed my test to no longer explicitly call addCustomElementChange on the physical and modelunderrepositorymodel, relying on the addCustomModelChange function to handle it.

My test started failing after I did that so then added your change in to call addCustomElementChange within addCustomModelChange and its now passing.

common/api/imodel-transformer.api.md
```diff
@@ -49,6 +73,12 @@ export class ChangedInstanceIds {
     element: ChangedInstanceOps;
     // (undocumented)
     font: ChangedInstanceOps;
+    // @beta
+    getCustomRelationshipDataFromId(id: Id64String): ChangedInstanceCustomRelationshipData | undefined;
```
JulijaRamoskiene, line 77: @aruniverse , @DanRod1999, can you advise me: 
`getCustomRelationshipDataFromId()` and `hasCustomRelationshipChanges()` are not meant to be called by third party code, those methods are supposed to be "internal". 
How to properly handle/mark such methods in 'imodel-transformer' package?I  feel we might want to remove them in future or implement in different way. Is it enough to mark them as 'beta'? Those will never be something that user should use.
I found one case where such internal method was marked as private and then called like this: 
https://github.com/iTwin/imodel-transformer/blob/4e4a7df0030ea7fba666c4b700dc6f4160074340/packages/transformer/src/IModelTransformer.ts#L3169

aruniverse: ah i misread, dan is right

packages/transformer/src/IModelExporter.ts
```diff
@@ -438,14 +448,21 @@ export class IModelExporter {
...
 
-    const startChangeset =
-      args && "startChangeset" in args ? args.startChangeset : undefined;
+    const isEmptyObject = (obj: object): boolean =>
+      Object.keys(obj).length === 0;
 
-    const initOpts: ExporterInitOptions = {
-      startChangeset: { id: startChangeset?.id },
-    };
+    let initOpts: ExporterInitOptions;
```
JulijaRamoskiene, line 454: I am unsure why this was added. I suspect Nick stumbled into some issue similar to the one I also reported: https://github.com/iTwin/imodel-transformer/issues/236

We should consider if this is really the right approach for this problem. My first instinct is that we should rather save processed version and use that as default value, instead of "process the current changeset of the source iModel being exported when startChangeset.id is undefined.". I am not sure if this change means only last (current) change set would be processed? In such case this could result in unpredictable behavior if there was more than 1 change set added since last transformation.

JulijaRamoskiene: I remember it now, this was related to:  https://github.com/iTwin/imodel-transformer/issues/241

PR #178, opened 2024-05-07
upgrade docs build tools; typescript; eslint-plugin

ben-polinsky:


ben-polinsky:
@nick4598 can you give this a look-over? Nothing too drastic

packages/transformer/src/IModelTransformer.ts
```diff
@@ -329,7 +329,7 @@ function mapId64<R>(
   } else {
     throw Error(
       [
-        `Id64 container '${idContainer}' is unsupported.`,
```
nick4598, line 332: Why did you have to change this to idContainer.id ?

nick4598: I would expect that the 'isRElatedElem' check
 `const isRelatedElem = (arg: any): arg is RelatedElement =>
    arg && typeof arg === "object" && "id" in arg;` 
    
would exclude the possibility of idContainer.id in that throw statement.

.github/workflows/ci.yml
```diff
@@ -117,17 +117,18 @@ jobs:
         const fs = require("fs");
         const workspacePkgJsonPath = "./package.json";
         const workspacePkgJson = require(workspacePkgJsonPath);
-        workspacePkgJson.pnpm.overrides = {
-          ...workspacePkgJson.pnpm.overrides,
```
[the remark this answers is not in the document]
ben-polinsky: Nope, I should have left that out.

packages/transformer/src/IModelTransformer.ts
ben-polinsky: No, apparently that type check isn't sufficient for `{ id: Id64String }`, only for `RelatedElement`. I can type the argument that way?

.github/workflows/ci.yml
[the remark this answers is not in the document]
ben-polinsky: Oh that is the old code...  
indecisive. We shouldn't be using overrides.. okay I'll bite..

packages/transformer/src/IModelTransformer.ts
nick4598: Im assuming if we ever make it there, we're logging an object which doesnt have an 'id' property defined on it. So not really sure what to do about that... I guess whatever logs the object seems best. I don't know when something gets logged as [Object object] or not though.

ben-polinsky: That's the root of these changes - eslint is complaining that it could be logged `[Object object]`. I'm using `JSON.stringify(idContainer)` now.

.github/workflows/ci.yml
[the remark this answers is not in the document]
ben-polinsky: Since I've removed the `pnpm` object, I'm spreading both that and `pnpm.overrides`. This whole deal should be revisited. As a comment above states, it is definitely hacky.

packages/transformer/src/IModelTransformer.ts
nick4598: That seems good to me. Thanks

.github/workflows/ci.yml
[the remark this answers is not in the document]
nick4598: thanks

nick4598 (approved):

PR #179, opened 2024-05-13
Support providing the reversesync and sync versions when unsafe-migrate is set.

nick4598:
Support providing the reversesync and sync versions when unsafe-migrate is set. This is to support svcs team and their use cases so they no longer have to have code surrounding this. A change we made to make syncDirection determined automatically by the transformer has made some of their patches no longer work.

I also ran into a bug when trying to test this which had 2 issues, the first being the hasElementChangedCache was getting aspect ids added to it. I believe it was ALWAYS supposed to be only element ids, but I made that mistake when converting to use Affan's changeset reader. The query that used to be used for the hasElementChangedCache had a where clause:  [`ic.ChangedInstance.ClassId IS (ONLY BisCore.Element)`](https://github.com/iTwin/imodel-transformer/commit/576295137997aa8911e5589f1b1377a19902221d#diff-d07eab136b812db5c475ddb91cbc589d5fd6e5943638c6e2028994d6f58fe80bR687)

The second part of the bug: I believe we should no longer be checking `if Id64.isValid(targetElementId)` when deciding if we should return early because an element has not changed. This seems to be leftover from a time when you needed the targetElementId to find the ESA which described the relationship between the source and target element. I believe at the time of Mike making his change to no longer search for ESAs he didn't want to remove targetElementId (so he changed it to _targetElementId), because that COULD be a breaking change if someone was overriding the hasElementChanged function.

The services team does not appear to be overriding the hasElementChanged function, and this change will be part of a 1.0.0 major release.

### DIFF
--- a/packages/transformer/src/IModelTransformer.ts
+++ b/packages/transformer/src/IModelTransformer.ts
@@ -247,6 +247,23 @@ export interface IModelTransformOptions {
    * @default "reject"
    */
   branchRelationshipDataBehavior?: "unsafe-migrate" | "reject";
+
+  /**
+   * The forward sync 'version' to set on the scoping ESA @see ExternalSourceAspectProps upon startup, if the version property on the scoping ESA is undefined or empty string.
+   * @note This option is not without risk! You must also set @see branchRelationshipDataBehavior to "unsafe-migrate".
+   * @note This value is ignored if the version property on the scoping ESA is NOT undefined or empty string.
+   * @default ""
+   */
+  unsafeSyncVersion?: string;
+
+  /**
+   * The reverse sync version to set on the scoping ESA @see TargetScopeProvenanceJsonProps upon startup, if the reverseSync property on the scoping ESA is undefined or empty string.
+   * @note This option is not without risk! You must also set @see branchRelationshipDataBehavior to "unsafe-migrate".
+   * @note This value is ignored if the reverseSyncVersion property on the scoping ESA is NOT undefined or empty string.
+   * @default ""
+   */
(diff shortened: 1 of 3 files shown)


packages/transformer/src/IModelTransformer.ts
```diff
@@ -247,6 +247,23 @@ export interface IModelTransformOptions {
...
+  unsafeSyncVersion?: string;
+
+  /**
+   * The reverse sync version to set on the scoping ESA @see TargetScopeProvenanceJsonProps upon startup, if the reverseSync property on the scoping ESA is undefined or empty string.
+   * @note This option is not without risk! You must also set @see branchRelationshipDataBehavior to "unsafe-migrate".
+   * @note This value is ignored if the reverseSyncVersion property on the scoping ESA is NOT undefined or empty string.
+   * @default ""
+   */
+  unsafeReverseSyncVersion?: string;
+
```
JulijaRamoskiene, line 266: Can we improve naming, to include info that this passed value will be used only if actual sync version is empty. My suggestion: `unsafeFallbackSyncVersion` and `unsafeFallbackReverseSyncVersion`.

nick4598: changed to unsafeFallbackSyncVersion

nick4598:
https://github.com/iTwin/itwinjs-core/pull/6643/files

Test is failing on 4.6.0 due to the above change

JulijaRamoskiene (dismissed):

JulijaRamoskiene (approved):

PR #181, opened 2024-05-14
Fix failing test which fails in 4.6.0 only due to a change in the way we handle deletions of DefinitionContainers

nick4598:
PR in 4.6.0 which made this change: https://github.com/iTwin/itwinjs-core/pull/6643/files

### DIFF
--- a/packages/transformer/src/IModelTransformer.ts
+++ b/packages/transformer/src/IModelTransformer.ts
@@ -2014,36 +2014,35 @@ export class IModelTransformer extends IModelExportHandler {
     // It is possible and apparently occasionally sensical to delete a model without deleting its underlying element.
     // - If only the model is deleted, [[initFromExternalSourceAspects]] will have already remapped the underlying element since it still exists.
     // - If both were deleted, [[remapDeletedSourceEntities]] will find and remap the deleted element making this operation valid
-    let sql: string;
-    if (this.hasDefinitionContainerDeletionFeature) {
-      sql = `
-      SELECT 1
-      FROM bis.DefinitionPartition
-      WHERE ECInstanceId=?
-      UNION
-      SELECT 1
-      FROM bis.DefinitionContainer
-      WHERE ECInstanceId=?
-    `;
-    } else {
-      sql = `
-      SELECT 1
-      FROM bis.DefinitionPartition
-      WHERE ECInstanceId=?
-    `;
-    }
     const targetModelId: Id64String =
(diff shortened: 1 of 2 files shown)


DanRod1999 (dismissed):

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1994,20 +2014,36 @@ export class IModelTransformer extends IModelExportHandler {
...
       const isDefinitionPartition = this.targetDb.withPreparedStatement(
-        `
-        SELECT 1
-        FROM bis.DefinitionPartition
-        WHERE ECInstanceId=?
-      `,
+        sql,
         (stmt) => {
           stmt.bindId(1, targetModelId);
+          if (this.hasDefinitionContainerDeletionFeature)
```
JulijaRamoskiene, line 2045: Since same id is bound twice, I suggest to use named parameter in your EC SQL instead:
```SQL
SELECT 1
FROM bis.DefinitionPartition
WHERE ECInstanceId=:targetModelId
UNION
SELECT 1
FROM bis.DefinitionContainer
WHERE ECInstanceId=:targetModelId
```
With named parameter you will not need `if` statement:
```typescript
stmt.bindId("targetModelId", targetModelId);
```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1994,20 +2014,36 @@ export class IModelTransformer extends IModelExportHandler {
     // It is possible and apparently occasionally sensical to delete a model without deleting its underlying element.
     // - If only the model is deleted, [[initFromExternalSourceAspects]] will have already remapped the underlying element since it still exists.
     // - If both were deleted, [[remapDeletedSourceEntities]] will find and remap the deleted element making this operation valid
+    let sql: string;
```
JulijaRamoskiene, line 2017: NitPick for readability: Suggest to move this initialization a bit lower right before it is used as: 

- It might not be required if  targetModelId is invalid.
- Currently it is a bit distracting, with this comment above.

JulijaRamoskiene (changes requested):

JulijaRamoskiene (dismissed):

packages/transformer/src/IModelTransformer.ts
nick4598: Good point, changed it.

JulijaRamoskiene (approved):

PR #190, opened 2024-07-10
Make processALl and processChanges private, expose a public process function and let people pass isSync to constructor of imodeltransformer

nick4598:


packages/transformer/src/IModelTransformer.ts
```diff
@@ -3225,13 +3238,50 @@ export class IModelTransformer extends IModelExportHandler {
...
+   *
+   * Export everything from the source iModel and import the transformed entities into the target iModel.
+   *
+   * Notes:
+   * - [[processSchemas]] is not called automatically since the target iModel may want a different collection of schemas.
+   *
+   * @throws if [[IModelTransformOptions.isSynchronization]] is true and [[ProcessChangesOptions]] is not provided.
+   */
+  public async process(options?: ProcessChangesOptions): Promise<void> {
+    if (!this._initialized) {
```
nick4598, line 3266: I ended up deciding to call initialize for someone if they didn't call it. Reason I leaned towards this is because there are amny process* functions, processFonts, processCodeSpecs etc. and I didn't want to make people call initialize for all of those. And in my opinion all process* functions should follow same behavior.

nick4598: I could be convinced otherwise I suppose, but I also changed the comment over initialize to instead not say "prefer to call process instead of calling initialize explicitly", but 'Callers may wish to explicitly call initialize if they need to execute code after initialize but before [[process]] is called.'

packages/transformer/src/IModelTransformer.ts
```diff
@@ -271,6 +271,12 @@ export interface IModelTransformOptions {
    * @default true
    */
   skipPropagateChangesToRootElements?: boolean;
+
+  /**
+   * Whether or not this transformation is a synchronization. This influences the behavior of @see [[IModelTransformer.process]].
+   * @default false
+   */
+  isSynchronization?: boolean;
```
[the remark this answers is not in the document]
JulijaRamoskiene: Maybe there is a reason why those options can't be passed to ctor.. checked what we are passing and.. turns out it is an empty object  
image

packages/transformer/src/IModelTransformer.ts
JulijaRamoskiene, line 3266: same `this._initialized` is validated in initialize(). Seems that we can skip it here

JulijaRamoskiene: Implementation is confusing to me from that point that user could pass processChangesOptions and processAll() would be still called. See my coment above about removing in favor of processChangesOptions.

JulijaRamoskiene (changes requested):

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
nick4598: `new IModelTransformer({processChangesOptions: { }})` could look strange, one might think that by passing an empty object you're opting out of processChanges, but really by passing { } you're telling me you want to run processChanges.

packages/transformer/src/IModelTransformer.ts
nick4598: Well actually initialize returns early if initialized is true, but this will call initialize for you if its currently false.

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
nick4598: I made a slight change, is it better worse or the same? 
image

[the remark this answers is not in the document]
nick4598: Okay i switched over to your suggestion

packages/transformer/src/IModelTransformer.ts
JulijaRamoskiene: My thinking is that `initialize ()` can be safely called as many times as you want, as it will initialize only once any way. No need to double check. Unless you have performance concerns to call `initialize ()`. Not sure if there is unit test for this behavior, but you could add one, to make sure it stays that way in future implementation:
`initialize () => should initialize only once when called multiple times`

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
JulijaRamoskiene: We looked at code with @mindaugasdirg  and (don't be mad), we have new suggestion! `Lets leave isSynchronization as it was before (property in transformaer options) and remove processChanges options instead`, as
- allowing user to pass access token is a path to time out issues that you helped us solve recently on our end. 
- startChangeset is also no longer needed (@mindaugasdirg  knows more)

Then we would achieve 2 things: 
- all configurations would be set in ctor
- no need for if's that check both `isSynchronization` and `processChangeOptions arguments` 

Hope it would bring us closer to having options that would be very straight forward to understand and configure.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -271,6 +271,12 @@ export interface IModelTransformOptions {
    * @default true
    */
   skipPropagateChangesToRootElements?: boolean;
+
+  /**
+   * Whether or not this transformation is a synchronization. This influences the behavior of @see [[IModelTransformer.process]].
+   * @default false
+   */
+  argsForProcessChanges?: ProcessChangesOptions;
 }
```
JulijaRamoskiene, line 280: I think it is time to remove ` isReverseSynchronization?: boolean;` option in `IModelTransformOptions`.

JulijaRamoskiene (changes requested):

packages/transformer/src/IModelTransformer.ts
nick4598: Oh I see what you're saying now, no if statement in process for the initialized var because initialize will return early anyways.

packages/transformer/src/IModelTransformer.ts
nick4598: removed

packages/transformer/src/IModelTransformer.ts
nick4598: removed the check in process

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
JulijaRamoskiene: I see that there were changes to add comment explaining that access token option should not be used, but my suggestion was to remove processChanges options (which would mean that accessToken option would be removed as well). Can you update me on latest decisions, and if you decided there is a reason to keep accessToken and startChangeset options? Thanks!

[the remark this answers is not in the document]
nick4598: image
These are the last messages I sent regarding my concerns on removing 'startChangeset' option.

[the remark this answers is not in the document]
JulijaRamoskiene: You are right, startChangeset option might be useful in some scenarios, so it makes sense to leave it. 

Suggestions for current implementation:

- Consider maybe 'ignoreMissingChangesetsInSynchronizations' could be dropped. It seems that if startChangesetIndex is set , then it already indicates that users intention is to ignore / skip missing change sets.  If you do not agree that this option can be removed, then it should be moved to `ProcessChangesOptions`.
- `unsafeFallbackSyncVersion` and `unsafeFallbackReverseSyncVersion` should be moved to `ProcessChangesOptions`
- I know you added comment to indicate that setting `accessToken` option is not safe, but maybe better approach would be to remove that option.

These suggestions are aimed to clean up and group all process changes options in one place.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -271,6 +263,12 @@ export interface IModelTransformOptions {
    * @default true
    */
   skipPropagateChangesToRootElements?: boolean;
+
+  /**
+   * Whether or not this transformation is a synchronization. This influences the behavior of @see [[IModelTransformer.process]].
```
JulijaRamoskiene, line 268: Update comment, it got outdated

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
nick4598: Do you think the transformer should error out explicitly if accesstoken is removed and no authorizationClient is defined on IModelHost?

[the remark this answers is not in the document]
JulijaRamoskiene: - It seems to me the same as when "access token is not passed and authorization client is not defined" in current implementation.  Do you know what error user would get right now? If error is not intuitive or clear, then it might be a good idea to check auth client configuration and provide users with better error. On the other hand, I expected `IModelHost.hubAccess.xxx()` to do the check, but it seems that it will just try to execute request with empty token (I might have missed something, haven't tried to actually execute it, only did a quick code check). 
-  Is AccessToken/Authorization client always required, or there are cases when transformation does not use access token?

packages/transformer/src/IModelTransformer.ts
nick4598: fixed

packages/test-app/src/Main.ts
```diff
@@ -576,7 +576,6 @@ void (async () => {
     if (processChanges) {
       assert(undefined !== args.sourceStartChangesetId);
       await Transformer.transformChanges(
-        await acquireAccessToken(),
```
nick4598, line 579: acquireAccessToken in this case seems to setup the authorizationClient on IModelHost. Don't we still need to call this somewhere?

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -3593,8 +3586,12 @@ describe("IModelTransformerHub", () => {
     ) => {
       transformer["_options"]["branchRelationshipDataBehavior"] =
         "unsafe-migrate";
-      transformer["_options"]["unsafeFallbackReverseSyncVersion"] = ";2";
-      transformer["_options"]["unsafeFallbackSyncVersion"] = ";3";
+      transformer["_options"][
+        "argsForProcessChanges"
+      ]!.unsafeFallbackReverseSyncVersion = ";2";
```
nick4598, line 3591: Can you use ["unsafeFallBackReverseSyncVersion"] instead? It gives us type safety on private vars

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -3593,8 +3586,12 @@ describe("IModelTransformerHub", () => {
...
+      transformer["_options"][
+        "argsForProcessChanges"
+      ]!.unsafeFallbackSyncVersion = ";3";
```
nick4598, line 3594: same as above

packages/test-app/src/Main.ts
derbynn: acquireAccessToken is also called by about 10 other places in this file and over there it initiates the auth client if it is not already created. I removed it from transformChanges because accessToken it was no longer used here after removing accessToken from processChange args

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
derbynn: Sounds good

packages/test-app/src/Main.ts
nick4598: As long as we're guaranteed to call it beforehand, then thats fine.

derbynn: Yeah it looks like there are a number of calls before this line

nick4598: But do they always occur? Or are they gated behind some conditionals?

derbynn: image
It looks they are gated behind conditionals but still called regardless and they follow the same format for other lines where they are called

nick4598: processChanges is true (causing Transformer.transformChanges to be called) when args.SourceStartChangesetIndex || args.sourceStartChangesetId, so seems safe

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
derbynn: @JulijaRamoskiene I'll be working with Nick. Can you please review my new changes to address the previous comments from above?  I have moved the appropriate options from IModelTransformoptions to processChangesOptions, log warning if there is no authorization client and remove accessToken as it can be generated if there is an authorization client provided.

JulijaRamoskiene (changes requested):

packages/transformer/src/IModelTransformer.ts
```diff
@@ -638,6 +635,10 @@ export class IModelTransformer extends IModelExportHandler {
       skipPropagateChangesToRootElements:
         options?.skipPropagateChangesToRootElements ?? true,
     };
+    // check if authorization client is defined
+    if (IModelHost.authorizationClient === undefined) {
+      Logger.logWarning(loggerCategory, "No authorization client provided");
```
JulijaRamoskiene, line 640: [nit pick] Warning could be more descriptive.  :
- Where auth client is not provided? Now it sounds like I need to provide it as ctor parameter. But in reality it is not set in iModelHost.

JulijaRamoskiene (changes requested):

packages/transformer/src/test/IModelTransformerUtils.ts
```diff
@@ -1540,9 +1541,12 @@ export class PhysicalModelConsolidator extends IModelTransformer {
   public constructor(
     sourceDb: IModelDb,
     targetDb: IModelDb,
-    targetModelId: Id64String
+    targetModelId: Id64String,
+    argsForProcessChanges?: ProcessChangesOptions
   ) {
-    super(sourceDb, targetDb);
+    super(sourceDb, targetDb, {
+      argsForProcessChanges,
```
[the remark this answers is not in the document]
derbynn: Before committing and pushing these changes, Prettier runs and formats the files. Adding the comma at the end is part of its formatting rules, so if it is removed, it will be added back

JulijaRamoskiene (approved):

PR #198, opened 2024-08-13
remove deprecated for 3.x, 0.1.x, 0.1.0

derbynn:


packages/transformer/src/IModelExporter.ts
```diff
@@ -427,16 +426,8 @@ export class IModelExporter {
    *       for consecutive calls
    */
   public async exportChanges(args?: ExportChangesOptions): Promise<void>;
```
nick4598, line 428: You dont need both exportChanges anymore

packages/transformer/src/IModelExporter.ts
```diff
@@ -448,13 +439,13 @@ export class IModelExporter {
...
 
-    const initOpts: ExporterInitOptions =
-      typeof accessTokenOrOpts === "object"
-        ? accessTokenOrOpts
-        : {
-            accessToken: accessTokenOrOpts,
-            startChangeset: { id: startChangesetId },
-          };
+    const startChangeset =
+      "startChangeset" in args! ? args.startChangeset : undefined;
```
nick4598, line 443: Instead of using ! can we just check if its defined before checking "startchangeset" in args

nick4598 (approved):

nick4598 (approved):

nick4598 (approved):

PR #207, opened 2024-09-11
Enable saving reverse synchronization version for process all transformations

derbynn:
- Add support to save reverse sync version needed for partial fork workflow


nick4598:
@derbynn It seems that we potentially save some pending indices when isProvenanceInitTransform is set to true. Could you log / determine what those are? I am curious if we need to be concerned with saving those at all when saveReverseVersion is true.

derbynn:
> @derbynn It seems that we potentially save some pending indices when isProvenanceInitTransform is set to true. Could you log / determine what those are? I am curious if we need to be concerned with saving those at all when saveReverseVersion is true.

@nick4598 It looks like we save those pending indices, when the `argsForProcessChanges` is defined or the `_startingChangesetIndices` is defined and `_isProvenanceInitTransform` is set to `true` -  https://github.com/iTwin/imodel-transformer/blob/36232826b26115fbc2ef78573e37fd5b72c5401a/packages/transformer/src/IModelTransformer.ts#L2269C3-L2272C8

For the actual pushing of indices into those pending indices array, the `_startingChangesetIndices` must be defined - https://github.com/iTwin/imodel-transformer/blob/36232826b26115fbc2ef78573e37fd5b72c5401a/packages/transformer/src/IModelTransformer.ts#L2311

For the `_startingChangesetIndices` to be defined, source db and target db must be a briefcase - https://github.com/iTwin/imodel-transformer/blob/36232826b26115fbc2ef78573e37fd5b72c5401a/packages/transformer/src/IModelTransformer.ts#L674

For tests in `IModelTransformerHub` that use briefcase db and `_isProvenanceInitTransform` is set to `true`, we save some indices in pendingReverseSyncChangesetIndices
image

So, maybe we might need to add a condition like this `(this._startingChangesetIndices && saveReverseVersion)`  to save the pending indices like we are doing here -  https://github.com/iTwin/imodel-transformer/blob/36232826b26115fbc2ef78573e37fd5b72c5401a/packages/transformer/src/IModelTransformer.ts#L2269C3-L2272C8 ?

nick4598 (dismissed):

nick4598:
@JulijaRamoskiene please take a look when you get a chance. I'm fairly confident that this will work for you guys but want to make sure its acceptable for you. I think your fixReverseSyncVersion code is still going to have to run, but now we will add reversesyncversion and a pendingreversesync cs index if you guys pass saveReverseVersion. (Pending reversesync cs index will be the one to skip the changeset pushed to the provenanceDb at the end of transformation)

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2234,6 +2239,11 @@ export class IModelTransformer extends IModelExportHandler {
     const sourceVersion = `${this.sourceDb.changeset.id};${this.sourceDb.changeset.index}`;
     const targetVersion = `${this.targetDb.changeset.id};${this.targetDb.changeset.index}`;
 
+    if (saveReverseVersion) {
+      this._targetScopeProvenanceProps.jsonProperties.reverseSyncVersion =
+        targetVersion;
+    }
```
JulijaRamoskiene, line 2245: Current implementation works, but I think there might be maintainability/functionality issues in future. 
My concerns are that: 
- as documentation says `saveReverseVersion` is needed only for processAll transformation (I think should say 'forward transformation' instead). In all other scenarios, reverseVersion is saved anyway. With current implementation, synch version is set to invalid value for second for reverse synch. This is not noticible as it  is fixed right away in code below: 
image

If that code is changed, value might get corrupted for processChanges in reverse direction.

Suggest to save reverse synch only for scenarios when it might be empty, and to set it to correct value only: 
image

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2221,10 +2221,15 @@ export class IModelTransformer extends IModelExportHandler {
...
    * without setting the `force` option to `true`
+   * The saveReverseVersion is added to save the reverse sync version for processAll transformations, when set to `true`, reverse sync version is saved
    */
-  public updateSynchronizationVersion({ force = false } = {}) {
+  public updateSynchronizationVersion({
+    force = false,
+    saveReverseVersion = false,
+  } = {}) {
     const notForcedAndHasNoChangesAndIsntProvenanceInit =
       !force &&
```
JulijaRamoskiene, line 2231: is it needed to check both `force` and 'saveReverseVersion' here? Maybe checking just `force` is enough here..?

packages/transformer/src/IModelTransformer.ts
nick4598: Definitely agree with you that this function is getting a bit out of control.
Wehn you say "With current implementation, synch version is set to invalid value for second for reverse synch. This is not noticible as it is fixed right away in code below:" could you elaborate? In what scenario would we pass saveReverseVersion during a reverseSync?

packages/transformer/src/IModelTransformer.ts
derbynn: if we expect to set force to true every time, we use saveReverseVersion, then maybe, the additional check might not be needed. I guess the point of setting force to true is so that we don't return early? So adding saveReverseVersion in that check might be unnecessary

derbynn: If someone is passing force does that mean they also want to saveReverseVersion in all cases?

packages/transformer/src/IModelTransformer.ts
derbynn: Clarified offline

packages/transformer/src/IModelTransformer.ts
derbynn: Clarified offline

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2220,25 +2220,23 @@ export class IModelTransformer extends IModelExportHandler {
    * You generally should not call this function yourself and use [[process]] with [[IModelTransformOptions.argsForProcessChanges]] provided instead.
    * It is public for unsupported use cases of custom synchronization transforms.
    * @note if [[IModelTransformOptions.argsForProcessChanges]] are not defined in this transformation, this will fail
-   * without setting the `force` option to `true`
+   * without setting the `initializeReverseSyncVersion` option to `true`
+   * The initializeReverseSyncVersion is added to save the reverse synchronization version for forward synchronization transformations. When set to `true`, reverse sync version is saved
```
nick4598, line 2224:
```suggestion
   * The initializeReverseSyncVersion is added to set the reverse synchronization version during a forward synchronization. When set to `true`, the reverse sync version is saved as the current changeset of the targetDb. Note that typically the reverseSyncVersion is saved as the last changeset merged from the branch into master. A user might decide to pass true to avoid having a reverseSyncVersion of "". A reverseSyncVersion of "" interferes with a fix for a now fixed bug. That fix cannot be easily removed, so we must support the initialization of the reverseSyncVersion
   ```

Comment (nick4598) [reply packages/transformer/src/IModelTransformer.ts:2224]: Not super in love with my suggestion but I would like to add more of the context for why this is necessaary.

File: packages/transformer/src/test/TestUtils/TimelineTestUtil.ts [hunk 4]
```diff
@@ -449,6 +449,9 @@ export async function runTimeline(
           wasSourceIModelCopiedToTarget: true,
         });
         await provenanceInserter.process();
+        provenanceInserter.updateSynchronizationVersion({
```
nick4598, line 452: How come you had to add this?

derbynn: Because this used to have the `_isProvenanceInitTransform` functionality that's why I added it there to save the first changeset.

packages/transformer/src/IModelTransformer.ts
nick4598: Probably it should also mention something about this only being needed for the initial transformation between a master and branch imodel

packages/transformer/src/test/TestUtils/TimelineTestUtil.ts
nick4598: But shouldnt that occur by passing the truth value of isProvenancwInitTrnasfomr to updateSyncVersion at the end of process?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2220,25 +2220,28 @@ export class IModelTransformer extends IModelExportHandler {
    * You generally should not call this function yourself and use [[process]] with [[IModelTransformOptions.argsForProcessChanges]] provided instead.
    * It is public for unsupported use cases of custom synchronization transforms.
    * @note if [[IModelTransformOptions.argsForProcessChanges]] are not defined in this transformation, this will fail
```
JulijaRamoskiene, line 2222: Documentation says: `if [[IModelTransformOptions.argsForProcessChanges]] are not defined in this transformation, this will fail without setting the `initializeReverseSyncVersion` option to `true``, but as I understand it will just return early (without updating synch version).

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2346,7 +2360,9 @@ export class IModelTransformer extends IModelExportHandler {
   // FIXME<MIKE>: is this necessary when manually using low level transform APIs? (document if so)
   private finalizeTransformation() {
     this.importer.finalize();
-    this.updateSynchronizationVersion();
+    this.updateSynchronizationVersion({
+      initializeReverseSyncVersion: this._isProvenanceInitTransform,
```
JulijaRamoskiene, line 2364: comma not needed

packages/transformer/src/IModelTransformer.ts
JulijaRamoskiene: @nick4598 which bug do you have in mind here? In our case, saving reverse synch version for forward synchronization is needed for partial fork workflow. If not saved, following transformation to synch new changes from fork to main iModel will fail, as transformer will process changes starting from first commit and will not remove deleted elements from main iModels. Send you more detailed documentation regarding this issue.
It is not needed to "+1" issue if you had that in mind.

JulijaRamoskiene (approved):

packages/transformer/src/IModelTransformer.ts
nick4598: I responded on teams my understanding of the situation, but Ill copy and paste it here as well:

Maybe Daniel could provide some of the context, if we were to not persist the reverseSyncVersion then it would be "". When this happens the core transformer treats the cachedSyncVersion (same as `_syncrhoinzationVersion` I think) for the transformation as index -1 and id "".
 
 
```
// NOTE: that we do NOT download the changesummary for the last transformed version, we want
    // to ignore those already processed changes
    const startChangesetIndexOrId =
      args.startChangeset?.index ??
      args.startChangeset?.id ??
      this._synchronizationVersion.index + 1;
```
this makes the syncVersion index start at 0. I believe there is no changeset 0, atleast Daniel showed that to me that we do not download a changeset 0. This seems in line with what I remember as well, that changeset 0 is really just the initial seed imodel before any changesets.
 
The core transformer stores pendingReverseSyncIndices of 1 after the initial transform (note that this did not previously happen and was a bug that you guys ended up fixing by adding 1 to the reverseSyncVersion ). Since the core transformer stores that pendingreverseSyncIndice it skips the changeset that we need to skip. Its the one that brought over all the hcanges from the partial fork. Since you guys fixed the bug by adding 1 to reverseSyncVersion you need the reverseSYncVersion to be ;initialized; so that you have ;0 stored there instead of "". Your fix is to split on ; and add 1 to the version. Since this is the way you guys fixed it and it has implications for your database, we of course need to continue to support this.
 
Thats my understanding of it, maybe Ive got it wrong. 
 
Personally I find it a bit confusing that we are initializing the reverseSyncVersino because its a bit opposite to what the documentatino of reverseSyncVersion is whicvh is that its the 'last changeset index reverse synced into master'. Thats the main reason I wanted as much documentation surrounding this.

packages/transformer/src/IModelTransformer.ts
nick4598: Might be prettiers doing? not sure

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2220,25 +2220,28 @@ export class IModelTransformer extends IModelExportHandler {
...
-  public updateSynchronizationVersion({ force = false } = {}) {
-    const notForcedAndHasNoChangesAndIsntProvenanceInit =
-      !force &&
-      this._sourceChangeDataState !== "has-changes" &&
-      !this._isProvenanceInitTransform;
-    if (notForcedAndHasNoChangesAndIsntProvenanceInit) return;
+   * without setting the `initializeReverseSyncVersion` option to `true`
+   * The `initializeReverseSyncVersion` is added to set the reverse synchronization version during a forward synchronization.
+   * When set to `true`, it saves the reverse sync version as the current changeset of the targetDb.
+   * Note that typically, the reverseSyncVersion is saved as the last changeset merged from the branch into master.
```
nick4598, line 2226:
```suggestion
   * Note that typically, the reverseSyncVersion is saved as the last changeset merged from the branch into master. This means passing initializeReverseSyncVersion could potentially overwrite a correct reverseSyncVersion which represents 'the last changeset merged from the branch into master', and should only be done during the first transformation between a master and branch imodel.
```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2220,25 +2220,28 @@ export class IModelTransformer extends IModelExportHandler {
...
+   * This is also necessary for the initial transformation between a master and branch imodel.
+   * Setting it to `true` avoids having a reverseSyncVersion of "", which interferes with a fix for a now resolved bug.
```
nick4598, line 2228:
```suggestion
   * Setting it to `true` avoids having a reverseSyncVersion of "" after the initial transformation, which interferes with a fix for a now resolved bug.
```

packages/transformer/src/IModelTransformer.ts
derbynn: That's correct. Prettier adds these commas
image

packages/transformer/src/IModelTransformer.ts
derbynn: added to documentation

packages/transformer/src/IModelTransformer.ts
derbynn: Thanks for the context, Nick. I agree that adding the new initializeReverseSyncVersion parameter makes it confusing since this new behavior does not reflect what is documented. We have added an additional comment to let users know of the effects of using it when it is not the first transformations -   
   * `Note that typically, the reverseSyncVersion is saved as the last changeset merged from the branch into master.`
   * `Setting initializeReverseSyncVersion to true could overwrite this correct reverseSyncVersion and should only be done during the first transformation between a master and branch iModel.`

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2212,33 +2212,38 @@ export class IModelTransformer extends IModelExportHandler {
...
-   * without setting the `force` option to `true`
+   * @note If [[IModelTransformOptions.argsForProcessChanges]] is not defined in this transformation, this function will return early without updating the sync version,
+   * unless the `initializeReverseSyncVersion` option is set to `true`
+   *
+   * The `initializeReverseSyncVersion` is added to set the reverse synchronization version during a forward synchronization.
+   * When set to `true`, it saves the reverse sync version as the current changeset of the targetDb. This is typically used for the first transformation between a master and branch iModel.
+   * Setting `initializeReverseSyncVersion` to `true` has the effect of making it so any changesets in the branch iModel at the time of the first transformation will be ignored during any future reverse synchronizations from the branch to the master iModel.
+   *
+   * Note that typically, the reverseSyncVersion is saved as the last changeset merged from the branch into master.
+   * Setting initializeReverseSyncVersion to true could overwrite this correct reverseSyncVersion and should only be done during the first transformation between a master and branch iModel.
```
nick4598, line 2231:
```suggestion
   * Setting initializeReverseSyncVersion to true during a forward transformation could overwrite this correct reverseSyncVersion and should only be done during the first transformation between a master and branch iModel.
```

nick4598 (approved):

packages/transformer/src/IModelTransformer.ts
nick4598: Going to add this for sake of completeness: but flaw in my logic as julija pointed out is that I assume target iModel starts off at changeset 0. But there are some cases where they need to push up some schemas in a separate changeset to the target iModle before running transformation. Setting up the reverse sync version in those cases is required to skip the changeset that brought over schema changes.

PR #209, opened 2024-09-26
Add a possibility to re-add deleted elements with preserved ids

derbynn:
- Perform early checks during exportElement to take advantage of `this.context.findTargetElementId` i.e. look for cases where element exists but has different id in target or element id in target is not available, as some other element uses that id and throw an error for those cases.
- Otherwise, update if we find the element in target db or insert if element is not in target db yet


packages/transformer/src/IModelTransformer.ts
```diff
@@ -1944,6 +1944,38 @@ export class IModelTransformer extends IModelExportHandler {
...
+        // Element found with different id
+        throw new Error(
+          `Element id(${sourceElement.id}) cannot be preserved. Found a different mapping(${elementInTargetId}) from source element`
+        );
+      }
+      // if we don't find mapping for source element in target(invalid) but an element with source id exists in target and it is not the same as source element
+      else if (
+        isInvalid &&
+        sourceInTargetElemProps?.federationGuid &&
+        sourceElemProps?.federationGuid !==
```
nick4598, line 1969: I dont think we need to compare federationGuids here. If `isInvalid` is already true, AND we found an element in the target with the sameId then we know they didnt have the same federationGuid otherwise further up in onExportElement would've found it when it calls `_queryElemIdByFedGuid`

NOTE: The above statement is actually not true because of preserveElementIdsForFiltering if statement at the top.
I wonder if we should scrap that if statement at the top and worry about it here? Then we can just check the targetElementId instead of calling
`const elementInTargetId = this.context.findTargetElementId(
        sourceElement.id
      );`

derbynn: To address your first statement, I found some tests that were failing and when I investigated. Though invalid, the element we found in the target was the same element as the source i.e. it had the same information as the source element. The only thing that was missing was the federation guid(I was not sure why getElementProps did not bring over the fed guid)

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1944,6 +1941,31 @@ export class IModelTransformer extends IModelExportHandler {
...
+        // Element found with different id
+        throw new Error(
+          `Element id(${sourceElement.id}) cannot be preserved. Found a different mapping(${targetElementId}) from source element`
+        );
+      }
+      // if we don't find mapping for source element in target(invalid) but another element with source id exists in target
+      else if (isInvalid && sourceInTargetElemProps) {
+        // Element id is already taken by another element
+        throw new Error(
+          `Element id(${sourceElement.id}) cannot be preserved. Found a mapping that is already used by another element(${sourceInTargetElemProps.id}) in target db`
```
nick4598, line 1960:
```suggestion
          `Element id(${sourceElement.id}) cannot be preserved. An unrelated element in the target already uses id: ${sourceElement.Id}`
```

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1944,6 +1941,31 @@ export class IModelTransformer extends IModelExportHandler {
       ? targetElementId
       : undefined;
 
+    if (this._options.preserveElementIdsForFiltering) {
+      const isInvalid = !Id64.isValid(targetElementId);
+      const sourceInTargetElemProps = this.targetDb.elements.tryGetElementProps(
```
nick4598, line 1946: Theres an opportunity for a slight optimization here, in that we dont need to call tryGetElementProps unless the first if statement is false.

packages/transformer/src/IModelImporter.ts
```diff
@@ -257,7 +258,14 @@ export class IModelImporter {
       ) {
         this.onUpdateElement(elementProps);
       } else {
-        this.onInsertElement(elementProps);
+        const doesElementExistInTarget = this.targetDb.elements.tryGetElement(
+          elementProps.id
+        );
+        if (doesElementExistInTarget) {
+          this.onUpdateElement(elementProps);
```
nick4598, line 265: I think we should also catch a potential error thrown by onupdateElement and do the duplicateCodeValueMap logic here

packages/transformer/src/IModelImporter.ts
```diff
@@ -257,7 +258,14 @@ export class IModelImporter {
       ) {
         this.onUpdateElement(elementProps);
       } else {
-        this.onInsertElement(elementProps);
+        const doesElementExistInTarget = this.targetDb.elements.tryGetElement(
```
nick4598, line 261: We definitely need a way to check this.. but I sort of want to avoid calling tryGetElement everytime. I will comment in imodeltransformer.ts for my suggestions: See comment on "targetElementProps.id = sourceElement.id"

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1944,6 +1941,31 @@ export class IModelTransformer extends IModelExportHandler {
...
+      // if we don't find mapping for source element in target(invalid) but another element with source id exists in target
+      else if (isInvalid && sourceInTargetElemProps) {
+        // Element id is already taken by another element
+        throw new Error(
+          `Element id(${sourceElement.id}) cannot be preserved. Found a mapping that is already used by another element(${sourceInTargetElemProps.id}) in target db`
+        );
+      }
+      // Element id in target is available to be remapped
+      else if (isInvalid) {
+        targetElementProps.id = sourceElement.id;
```
nick4598, line 1965: Not sure what the best way to do this but two options I see:

1)Add optional arg to importElement that tells importer we were in this if statement. ( if we're in this if statement we know we can call insertElement I think? Similarly, if our remapping is not invalid AND the ids are equal then we know we can call update I believe.

2) add public function to importElement: `markElementAsUpdate`. This adds the ID of the element to a set. Importer checks if the ID is in the set, if it is it removes it from the set and calls update. If its not then we know to call insert. 

I sort've lean towards the 2nd option just because the first would be an arg thats specific to preserveElementIdsForFiltering, and Im not sure how it would affect any subclasses of the importer elsewhere, but it does feel more hacky.

packages/transformer/src/IModelImporter.ts
derbynn: added a function to catch this potential errors and do the duplicateCodeValueMap logic so that we can add to both places and avoid duplication

packages/transformer/src/IModelImporter.ts
derbynn: added a new option `fromTransformation` that will be used to indicate whether importer class comes from transformation. If it is false, we can call tryGetElement to see whether an element exists in target. Otherwise, if importer comes from transformation, we can use the set to keep track of the elements to update during transformation

packages/transformer/src/IModelImporter.ts
```diff
@@ -63,6 +63,11 @@ export interface IModelImportOptions {
    * @see [IModelImporter Options]($docs/learning/transformer/index.md#IModelImporter)
    */
   autoExtendProjectExtents?: boolean | { excludeOutliers: boolean };
+  /**
+   * Indicates whether the importer is derived from a transformation process.
+   * If `true`, the importer originates from a transformation.
```
nick4598, line 68: You say the importer originates from a transformation, but I think that could be confusing because we could construct an instance of the importer and pass it to the transformer and `fromTransformation` would still be true right?

packages/transformer/src/IModelImporter.ts
```diff
@@ -109,6 +114,12 @@ export class IModelImporter {
    */
   public readonly doNotUpdateElementIds = new Set<Id64String>([]);
 
+  /** The set of elements that are added during the export of element during transformation to keep track of which elements to update during an import
+   * Defaults to an empty set.
+   * @note This is used when preserveElementIdsForFiltering is set to true
```
nick4598, line 119: I think you should mention that this is used as an optimization during the preserveElementIdsForFiltering behavior and that in normal cases where that option is false, the importer knows whether to insert or update based off of whether the ID is defined on the elementProps. The transformer already knows whether the element exists or not by time of calling importElement, so transformer pushes to the set with `markElementAsUpdate`. We can't do that in the preserveIds case because the ID should always be set whether the element exists or not. 

Maybe also mention that this set should stay small as we push to it and immediately remove from it? Or something like that?

packages/transformer/src/IModelImporter.ts
```diff
@@ -109,6 +114,12 @@ export class IModelImporter {
...
+   */
+  public elementsToUpdate = new Set<Id64String>([]);
```
nick4598, line 121: Can be private right?

packages/transformer/src/IModelImporter.ts
derbynn: Yeah, I can see how that will be confusing.

I was looking at it somehow wrongly, from the point of view where inside the transformation code in the constructor, the importer instance is created if it is not passed into the transformer but we could also pass it into the transformer

packages/transformer/src/IModelImporter.ts
```diff
@@ -63,6 +63,11 @@ export interface IModelImportOptions {
    * @see [IModelImporter Options]($docs/learning/transformer/index.md#IModelImporter)
    */
   autoExtendProjectExtents?: boolean | { excludeOutliers: boolean };
+  /**
+   * Indicates whether the importer is used in a transformation process.
+   * If `true`, the importer is passed into the transformer.
```
nick4598, line 68:
```suggestion
   * If `true`, the importer was passed into the transformer, or the transformer constructed this instance of the importer
```

packages/transformer/src/IModelImporter.ts
```diff
@@ -103,6 +108,20 @@ export class IModelImporter {
    */
   private _duplicateCodeValueMap: Map<Id64String, string>;
 
+  /**
+   * The set of elements added during export of an element in transformation to track which elements to update during import.
```
nick4598, line 112:
```suggestion
   * A set of elementIds that the transformer adds to while exporting elements to indicate that the element already exists in the target.
```

nick4598 (commented):
Do you know if all 4 cases described in the work item are covered by our tests?

packages/transformer/src/IModelImporter.ts
```diff
@@ -103,6 +108,20 @@ export class IModelImporter {
...
+   * Defaults to an empty set.
+   * @note
+   *
+   * This is used as an optimization when `preserveElementIdsForFiltering` is set to `true`
+   * In normal cases where this option set to `false`,
+   * the importer determines whether to insert or update based off of whether the ID is defined on `elementProps`.
+   * However, with `preserveElementIdsForFiltering` set to `true`, IDs are always set, so we can't determine insert/update like the normal case.
+   * The transformer already knows if an element exists or not by the time `importElement` is called and pushes to this set with `markElementAsUpdate`.
+   *
```
nick4598, line 121:
```suggestion
   * @note This set should stay small, as right after the transformer pushes to it, the importer will remove from the set.
```

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -2655,6 +2655,262 @@ describe("IModelTransformer", () => {
...
+    // Calling process() for second time with option to preserve elements in hopes of restoring deleted element
+    const secondTransformer = new IModelTransformer(sourceDb, targetDb, {
+      preserveElementIdsForFiltering: true,
+    });
+    await secondTransformer.process(); // should not throw error: duplicate code (65547) and should re-add deleted element
+    targetDb.saveChanges();
+
+    sourceContent = await getAllElementsInvariants(sourceDb);
+    targetContent = await getAllElementsInvariants(targetDb);
+    expect(targetContent).to.deep.equal(sourceContent);
```
nick4598, line 2721: can we also assert explicitly that the deleted subject 1 from target is back and has the same ID and same content?

nick4598 (approved):

nick4598 (dismissed):

packages/transformer/src/IModelImporter.ts
```diff
@@ -229,6 +259,25 @@ export class IModelImporter {
 
   /** Import the specified ElementProps (either as an insert or an update) into the target iModel. */
   public importElement(elementProps: ElementProps): Id64String {
+    const tryUpdateElement = (elemProps: ElementProps): void => {
```
JulijaRamoskiene, line 262: [NitPick] Maybe its is just preference, but methods are becoming very long.
I think this part (as many others) would benefit if code was split into smaller separate functions e.g. in this case, why not move this to private method, instead of having it as local variable.

JulijaRamoskiene (changes requested):

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -2655,6 +2655,271 @@ describe("IModelTransformer", () => {
...
+    // Calling process() for second time with option to preserve elements in hopes of restoring deleted element
+    const secondTransformer = new IModelTransformer(sourceDb, targetDb, {
+      preserveElementIdsForFiltering: true,
+    });
+    await secondTransformer.process(); // should not throw error: duplicate code (65547) and should re-add deleted element
+    targetDb.saveChanges();
+
+    // verify that deleted element in target is added back - redundant check for explicitness
+    const sourceElementJSON = sourceDb.elements
+      .tryGetElement<Subject>(targetSubjectId1!)
```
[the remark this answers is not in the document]
derbynn: Added

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1944,6 +1942,34 @@ export class IModelTransformer extends IModelExportHandler {
       ? targetElementId
       : undefined;
 
+    if (this._options.preserveElementIdsForFiltering) {
+      const isInvalid = !Id64.isValid(targetElementId);
```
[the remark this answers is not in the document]
derbynn: Added

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -2655,6 +2655,271 @@ describe("IModelTransformer", () => {
...
+
+    sourceContent = await getAllElementsInvariants(sourceDb);
+    targetContent = await getAllElementsInvariants(targetDb);
+    expect(targetContent).to.deep.equal(sourceContent);
+
+    sourceDb.close();
+    targetDb.close();
+  });
+
+  it("process() with preserveElementIdsForFiltering set to true should not throw when called on 2 identical iModels", async () => {
```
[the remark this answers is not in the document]
derbynn: Added

packages/transformer/src/IModelImporter.ts
```diff
@@ -155,6 +175,16 @@ export class IModelImporter {
...
+  /**
+   * Marks an element so that it can be updated during import when [[IModelImportOptions.preserveElementIdsForFiltering]] is set to true
+   */
+  public markElementAsUpdate(elementId: Id64String) {
+    if (this._rootElementIds.has(elementId)) {
+      return;
+    }
+    this._elementsToUpdate.add(elementId);
+  }
+
```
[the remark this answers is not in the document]
derbynn: - Thanks for the suggestions
- Removed `fromTransformation` and `preserveElementIdsForFiltering` importer options.
- Updated the names to `elementIdsToUpdateForPreserveId` and `markElementToUpdateForPreserveId` to better reflect their functions (let me know if these names work):
  - `elementIdsToUpdateForPreserveId` tracks the element IDs to update when `preserveElementIdsForFiltering` is set to true.
  - `markElementToUpdateForPreserveId` is called when `targetElementId` is valid (indicating an update) during export in transformation, adding the element to the `elementIdsToUpdateForPreserveId` set.
- We don't need to track elements to insert, as we can determine whether to preserve an element ID during insert with the `forceUseId` flag. If the set is defined, it indicates that the ID should be preserved. The set is initialized when the `preserveElementIdsForFiltering` option is set to true, if it is undefined.

image

packages/transformer/src/IModelImporter.ts
derbynn: Added

packages/transformer/src/IModelImporter.ts
```diff
@@ -110,17 +103,17 @@ export class IModelImporter {
 
   /**
    * A set of elementIds that the transformer adds to while exporting elements to indicate that the element already exists in the target.
-   * Defaults to an empty set.
+   * Defaults to undefined.
    * @note
    *
-   * This is used as an optimization when `preserveElementIdsForFiltering` is set to `true`
+   * This is used as an optimization when the transformer `preserveElementIdsForFiltering` option is set to `true`
```
nick4598, line 109: maybe we should reference imodeltransformoptions here?

packages/transformer/src/IModelImporter.ts
```diff
@@ -176,13 +166,19 @@ export class IModelImporter {
   }
 
   /**
-   * Marks an element so that it can be updated during import when [[IModelImportOptions.preserveElementIdsForFiltering]] is set to true
+   * Marks an element so that it can be updated during import when the [[IModelTransformOptions.preserveElementIdsForFiltering]] is set to true.
```
nick4598, line 169:
```suggestion
   * Marks an element so that it can be updated during import when [[IModelTransformOptions.preserveElementIdsForFiltering]] is set to true.
```

packages/transformer/src/IModelImporter.ts
```diff
@@ -300,30 +315,22 @@ export class IModelImporter {
...
-        const shouldUpdateElement =
-          !this.options.fromTransformation &&
-          this.targetDb.elements.tryGetElement(elementProps.id);
-        if (
-          this._elementsToUpdate.has(elementProps.id) ||
-          shouldUpdateElement
-        ) {
-          tryUpdateElement(elementProps);
-          this._elementsToUpdate.delete(elementProps.id);
+        if (this.elementIdsToUpdateForPreserveId?.has(elementProps.id)) {
```
nick4598, line 324: It feels weird to me to have `?.has` . I'm guessing this is necessary because the compiler doesn't know that shouldPreserveElementIdsForFiltering() guarantees that the set is defined?

Maybe the shouldPreserveElementIdsForFiltering function isn't necessary and we should just check the set directly?

Kind of a nitpick I guess

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1943,17 +1932,21 @@ export class IModelTransformer extends IModelExportHandler {
       : undefined;
 
     if (this._options.preserveElementIdsForFiltering) {
-      const isInvalid = !Id64.isValid(targetElementId);
+      // initialize set if undefined to indicate in importer that preserveElementIdsForFiltering option is true
+      if (!this.importer.elementIdsToUpdateForPreserveId) {
```
nick4598, line 1936: How do you feel about doing this right after we construct an instance of the importer? (or if we're given one too)

packages/transformer/src/IModelImporter.ts
nick4598: I guess you felt the need to do this because we need to check it in onInsertElement as well. Maybe its fine then, feel free to resolve

packages/transformer/src/IModelTransformer.ts
derbynn: yeah, that will a better place to add it

packages/transformer/src/IModelImporter.ts
derbynn: Yeah, the ?. has is there because the compiler can't tell that shouldPreserveElementIdsForFiltering() guarantees that the set is defined

packages/transformer/src/IModelImporter.ts
derbynn: Added

packages/transformer/src/IModelTransformer.ts
derbynn: Added

JulijaRamoskiene (changes requested):

packages/transformer/src/IModelImporter.ts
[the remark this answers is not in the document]
JulijaRamoskiene: - I think it is great that `fromTransformation` is gone
- I see that `preserveElementIdsForFiltering` was re-implemented as `enablePreserveElementIds()`. My initial assumption was that we can get rid of `preserveElementIdsForFiltering ` but since it is not true, can we revert this part to have `preserveElementIdsForFiltering` instead of `enablePreserveElementIds` as it was before. I think there is not enough value in this change, to justify this breaking change to user.

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -2655,6 +2655,322 @@ describe("IModelTransformer", () => {
...
+
+  it("process() with preserveElementIdsForFiltering set to true should update the element properties if element exists with desired id in target", async () => {
+    const sourceDbPath = IModelTransformerTestUtils.prepareOutputFile(
+      "IModelTransformer",
+      "PreserveIdOnTestModel-Source.bim"
+    );
+    const sourceDb = SnapshotDb.createEmpty(sourceDbPath, {
+      rootSubject: { name: "iModelA" },
+    });
+    Subject.insert(sourceDb, IModel.rootSubjectId, "Subject1");
```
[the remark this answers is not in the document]
derbynn: For this test, it seems useful to save. But for others tests with preserveIdForFiltering set to true, I think it might be useful to query in the target to see if it exists.

JulijaRamoskiene (approved):

nick4598 (approved):

Issue #214, opened 2024-10-11
require extract-api on PRs and update our current common/api/core-transformer.api.md


nick4598:
We should also make sure that https://www.itwinjs.org/reference/imodel-transformer/ updates to reflect any changes that we've made to our current common/api/core-transformer.api.md

PR #216, opened 2024-10-14
add extract-api.yaml

DanRod1999:
resolves: https://github.com/iTwin/imodel-transformer/issues/214


.github/workflows/extract-api.yaml
```diff
@@ -0,0 +1,73 @@
...
+        run: |
+          git config --local user.email imodeljs-admin@users.noreply.github.com
+          git config --local user.name imodeljs-admin
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 7.27.0
+
+      - name: Use Node.js 18
```
MichaelSwigerAtBentley, line 38: Node 18 EoL is coming up, maybe we should add an issue to update to 22

MichaelSwigerAtBentley (dismissed):

.github/workflows/extract-api.yaml
DanRod1999: good call. don't think this is the only occurrence. I'll check and maybe open a new pr just for the sake of easier tracking

MichaelSwigerAtBentley:
Adding my review, Hold on merging until someone else approves as well

DanRod1999:
> Adding my review, Hold on merging until someone else approves as well

yeah going to wait for Nick as well since this required some file name changes that may or may not be necessary

.github/workflows/extract-api.yaml
MichaelSwigerAtBentley: Yes,18 is used across the board it seems, should be a new PR. Just wanted to bring attention to it. Resolving

nick4598 (approved):

Issue #191, opened 2024-07-15
in packages/transformer/src/transformer.ts, make the error message for mismatching peer dependencies better.


nick4598:
Recently I was approached with a question about why a line of code like 

`import {IModelTransformer} from @itwin/imodel-transformer` was enough to crash a program on startup 
I didn't know the answer unfortunately, but eventually that person was able to determine that the crash was [coming from the check for matching peer deps.](https://github.com/iTwin/imodel-transformer/blob/0f8e8a5b3a8ff3812d37621634e6438ac1ef49ec/packages/transformer/src/transformer.ts#L31C1-L34C134)

This person had to find the error in the debugger, which seems unfortunate and I'd rather them be able to see this error immediately. I'm not sure if this is possible or not though.

### Action Items
- [ ] Determine if its possible to have this error message surfaced without going into a debugger.
  - This should be reproducible with a simple test package that imports from the imodel-transformer with a dev version of core-backend.
  - At a minimum this should be explicitly called out in our README taking care to mention special env vars in transformer.ts like TRANSFORMER_NO_STRICT_DEP_CHECK
- [ ] Don't throw the error if someone is using a dev package. Some check like semver.satisfies(4.7.0-dev.12, ^4.6.0) I believe will return false (this should be proven). We should probably consider this as acceptable and not throw the error.

MichaelBelousov:
This was added because this library relied (and I imagine still does) on internal (sometimes native addon) APIs.

You can get much harder to understand errors without this probably.

Yes it's a false positive in a lot of cases, especially when you opt to use "-dev" versions. 

Yes you're right that semver.satisfies never returns true when comparing differently tagged versions. 

You can probably clear the dev version mostly safely considering how it's used, but it is incorrect in some real cases.

But why is the user swallowing the error such that it is so hard to find in the first place? 

Any logger they're using should be printing uncaught errors, which is the default behavior anyway.

If the user didn't understand the error once found, maybe that also needs an improvement. (E.g. mention that -dev versions aren't considered (unless you change that))

nick4598:
You're right the library does still rely on some internal APIs, there has been some movement in the itwinjs core repo in the direction of [making it harder to use internal APIs](https://github.com/iTwin/itwinjs-core/pull/6945).

@ashar-bentley any ideas regarding "But why is the user swallowing the error such that it is so hard to find in the first place?"

PR #220, opened 2024-10-23
Improve error messages for mismatching peer dependencies 

derbynn:
- Include dev versions that still satisfy semver
- Address #191


packages/transformer/README.md
```diff
@@ -31,3 +31,20 @@ require("https").get("https://registry.npmjs.org/@itwin/imodel-transformer", r=>
 }))
 EOF
 ```
+
+## Environment Variables
+
+### TRANSFORMER_NO_STRICT_DEP_CHECK
+
+Set this variable to `1` to disable strict dependency checks in `packages/transformer/src/transformer.ts`.
```
nick4598, line 39:
```suggestion
Set this variable to `1` to disable strict dependency checks in `packages/transformer/src/transformer.ts`. The strict dependency check validates that the version of @itwinjs/core-backend provided falls in the range defined by the peerDependencies of the transformer package.
```

nick4598 (dismissed):

nick4598:
Also were you seeing the error message thrown in your test app without any extra work needed to be done? It just showed up?

nick4598 (approved):

PR #222, opened 2024-11-22
Add test which introduces a EntityClass that has a CA 'QueryView' and a navigation property to exercise ECReferenceTypesCache

nick4598:


packages/transformer/src/ECReferenceTypesCache.ts
```diff
@@ -221,6 +222,38 @@ export class ECReferenceTypesCache {
...
+
+    for (const [customAttributeName, _customAttribute] of targetCAs) {
+      if (
+        customAttributeName.toLowerCase() === "ecdbmap.queryview" &&
+        targetType === undefined
+      ) {
+        // I've essentially captured the case that will cause us problems, but now the question is what to do about it?
+        Logger.logInfo(
+          TransformerLoggerCategory.IModelTransformer,
+          `targetClass: ${targetClass.schema.name}:${targetClass.name} has customAttribute which is QueryView and no sourceType in the rootClassToRefType map`
```
nick4598, line 251:
```suggestion
          `targetClass: ${targetClass.schema.name}:${targetClass.name} has customAttribute which is QueryView and no type found in the rootClassToRefType map`
```

packages/transformer/src/ECReferenceTypesCache.ts
```diff
@@ -221,6 +222,38 @@ export class ECReferenceTypesCache {
...
+    for (const [customAttributeName, _customAttribute] of sourceCAs) {
+      // I've essentially captured the case that will cause us problems, but now the question is what to do about it?
+      if (
+        customAttributeName.toLowerCase() === "ecdbmap.queryview" &&
+        sourceType === undefined
+      ) {
+        // no sourceType means the rootBisClass.name is not in the bisRootClassToRefType map.
+        Logger.logInfo(
+          TransformerLoggerCategory.IModelTransformer,
+          `sourceClass: ${sourceClass.schema.name}:${sourceClass.name} has customAttribute which is QueryView and no sourceType in the rootClassToRefType map`
```
nick4598, line 237:
```suggestion
          `sourceClass: ${sourceClass.schema.name}:${sourceClass.name} has customAttribute which is QueryView and no type found in the rootClassToRefType map`
```

packages/transformer/src/ECReferenceTypesCache.ts
```diff
@@ -221,6 +222,42 @@ export class ECReferenceTypesCache {
...
       ECReferenceTypesCache.bisRootClassToRefType[targetRootBisClass.name];
+
+    const sourceCAs = sourceClass.getCustomAttributesSync();
+    const targetCAs = targetClass.getCustomAttributesSync();
+    for (const [customAttributeName, _customAttribute] of sourceCAs) {
+      /*
+       * Since queryViews are typically associated with abstract classes and have no specific table in the database, we can ignore them.
+       */
+      if (
+        customAttributeName.toLowerCase() === "ecdbmap.queryview" &&
```
nick4598, line 233: Worried about potential performance implications, if I am checking the customattributes of every relationshipClass for the queryView attribute, will that be slow? Potential optimization is to only check if the sourceType or targetType are undefined? 

I imagine this is probably quite a small %age of any given transformer run, so maybe it doesn't matter?

nick4598: Started timing it: Info | core-backend.IModelTransformer | Checked 1444 classes for QueryView custom attribute. This took 4ms.

Its maybe overkill.

packages/transformer/src/ECReferenceTypesCache.ts
```diff
@@ -221,6 +274,24 @@ export class ECReferenceTypesCache {
       ECReferenceTypesCache.bisRootClassToRefType[sourceRootBisClass.name];
     const targetType =
```
nick4598, line 275: @ColinKerr What is the recommended way to tell if a class is abstract or not?
Right now if targetType is undefined from calling `ECReferenceTypesCache.bisRootClassToRefType[targetRootBisClass.name];` That seems like a pretty good indicator that something is abstract, because it didn't have any matches for Element, Model, ElementAspect, or Relationship. But I do worry that it isn't foolproof. What if something else went wrong to cause us to get undefined? Right now the code assumes it is an error to get undefined and throws.

```
private static bisRootClassToRefType: Record<
    string,
    ConcreteEntityTypes | undefined
  > = {
    /* eslint-disable quote-props, @typescript-eslint/naming-convention */
    Element: ConcreteEntityTypes.Element,
    Model: ConcreteEntityTypes.Model,
    ElementAspect: ConcreteEntityTypes.ElementAspect,
    ElementRefersToElements: ConcreteEntityTypes.Relationship,
    ElementDrivesElement: ConcreteEntityTypes.Relationship,
    // code spec is technically a potential root class but it is ignored currently
    // see [ConcreteEntityTypes]($common)
    /* eslint-enable quote-props, @typescript-eslint/naming-convention */
  };
```

.github/workflows/ci.yml
```diff
@@ -19,12 +19,12 @@ jobs:
     - name: Install pnpm
       uses: pnpm/action-setup@v4
       with:
-        version: 7.27.0
+        version: 9.14.2
 
     - name: Use Node.js 18
```
nick4598, line 24:
```suggestion
    - name: Use Node.js 20
```

.github/workflows/ci.yml
```diff
@@ -102,12 +102,12 @@ jobs:
...
```
nick4598, line 107:
```suggestion
    - name: Use Node.js 20
```

PR #223, opened 2024-12-09
Supporting changing exclude criteria alongside entity recreation

nick4598:
If an entity is detected as recreated and the recreated entity is excluded by the exporter, then we should propagate the original delete over to the target.

### DIFF
--- a/packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
+++ b/packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
@@ -1302,7 +1302,11 @@ describe("IModelTransformerHub", () => {
         master: {
           sync: [
             "branch1",
-            { initTransformer: setForceOldRelationshipProvenanceMethod },
+            {
+              init: {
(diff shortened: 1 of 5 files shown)


packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -2975,6 +2983,97 @@ describe("IModelTransformerHub", () => {
     }
   });
 
+  for (const excludeBeforeTransformerInitialize of [true, false]) {
+    it(`should be able to handle entity recreation and a switching of exclude criteria ${excludeBeforeTransformerInitialize ? "before transformer initialize" : "after transformer initialize"}`, async () => {
```
nick4598, line 2987: I will probably also setup a parent/child element recreation situation and make sure that a delete of the parent propagated to the target also deletes the child. Not sure if thats necessary or not since it feels like mostly testing deletes work as intended.

nick4598:
I am making an assumption that we should propagate this delete. It feels correct to me, but please comment if you disagree

PR #228, opened 2025-01-07
Add the example code-snippet from itwinjs-core to this repo instead

nick4598:


packages/transformer/src/test/example-code/IModelBranchingOperations.test.ts
```diff
@@ -0,0 +1,317 @@
...
+  arbitraryEdit.editCounter++;
+  db.elements.insertElement(physicalObjectProps);
+  db.saveChanges();
+  await db.pushChanges({
+    accessToken: myAccessToken,
+    description,
+  });
+}
+
+namespace arbitraryEdit {
```
MichaelSwigerAtBentley, line 268: Curious as to why this counter is outside of the function, is it expected to be referenced elsewhere?

nick4598: My guess is its to achieve a static variable, which I suppose aren't supported in javascript (I.e. just writing static in front of a var within a function)

docs/learning/transformer/index.md
ben-polinsky, line 1: On line 74 - that method has moved to transformer, so you need `The [IModelCloneContext]($transformer)` to get the docs build passing

PR #234, opened 2025-02-10
iTwin.js v5

### LINKED ISSUE #262: `ViewDefinition2d` element is processed before its "baseModel" element is processed
It appears that for elements that are of the class that extend [`ViewDefinition2d`](https://www.itwinjs.org/reference/core-backend/viewdefinitions/viewdefinition2d/) are processed before their "baseModel"s. This is resulting in the transformed element to have `baseModelId` to be set as `0`.

I ran into this issue when running tests but it is likely that this could happen in the scenario of a real app as well. In order to reproduce:
- run [this test block](https://github.com/iTwin/imodel-transformer/blob/6a2dabe4a55d70814682723839850a3186af7d3d/packages/transformer/src/test/standalone/IModelTransformer.test.ts#L209) using debugger 
- [optional] you can break on [this line](https://github.com/iTwin/imodel-transformer/blob/6a2dabe4a55d70814682723839850a3186af7d3d/packages/transformer/src/test/standalone/IModelTransformer.test.ts#L265) where the transformer code is being called to verify it gets there
- the transformer processor ends up eventually calling [`cloneElement()`](https://github.com/iTwin/imodel-transformer/blob/6a2dabe4a55d70814682723839850a3186af7d3d/packages/transformer/src/IModelCloneContext.ts#L57) method here with the element that needs to be transformed. Specifically this method gets called with `sourceElement.id` as `'0x39'`, `sourceElement.baseModelId` as `'0x1a'` and `sourceElement.classFullName` as `'BisCore:DrawingViewDefinition'`. Note that there are also references to other elements in sourceElement object - `displayStyleId`, `categorySelectorId`, `model`, etc.
- step over the call in that line and examine contents of `targetElementProps`
    - `baseModelId`'s value is set to `'0'`
    - all other references to other elements - `displayStyleId`, `categorySelectorId`, `model`, etc. -  are valid guids

I realized that the element with the guid that `sourceElement.baseModelId` is not being processed before the sourceElement itself. However, all other props - `displayStyleId`, `categorySelectorId`, `model`, etc. - elements are processed before `sourceElement`. My hypothesis is that because of this order of processing, `cloneElement()` call can't map the `baseModelId` correctly in target iModel so sets it to `'0'` and returns it. @DanRod1999 checked native code briefly and he is in agreement with this hypothesis. The "baseModel" is inserted into the target iModel only after the ViewDefinition2d element has been inserted.

Interestingly, if you set a breakpoint [here in the test](https://github.com/iTwin/imodel-transformer/blob/6a2dabe4a55d70814682723839850a3186af7d3d/packages/transformer/src/test/standalone/IModelTransformer.test.ts#L266) - after the transform process has completed. Then in the debug console run `targetDb.elements.getElement("0x2e")`(the source element's id after transformation becomes `0x2e`), you can see that `baseModelId =
'0x45'` which is correct and not the `'0'` we saw earlier. Somewhere during the transformer's processing logic, it seems to fix the baseModelId discrepancy automatically. [I am still investigating how the auto-fix is taking place and will update this thread with findings]

### Sounds like everything works out, what's the problem??

There are 2 main issues with current state of things: 

From the explanation above we have seen that there is at least one instance where the ViewDefinition2d element is being processed before its "baseModel". By the time the entire transform "processing" has been completed, the `baseModelId` is correctly set but since we don't know what is causing this reconciliation, we can't be certain the auto-fixing occurs in **every** instance.

I discovered this issue initially upgrading iTwin.JS core package versions to 5.x. More context - the transformer after cloning elements eventually ends up [inserting the cloned element](https://github.com/iTwin/imodel-transformer/blob/6a2dabe4a55d70814682723839850a3186af7d3d/packages/transformer/src/test/IModelTransformerUtils.ts#L2071). In the inserted element props the `baseModelId` is supplied as `'0'`. The same element is [immediately fetched](https://github.com/iTwin/imodel-transformer/blob/6a2dabe4a55d70814682723839850a3186af7d3d/packages/transformer/src/test/IModelTransformerUtils.ts#L2072) but the get operation fails because baseModelId is not defined and throws. More details on that can be found here: https://github.com/iTwin/itwinjs-core/issues/8150. The consensus reached in that linked issue was that core-backend should throw an error if an insert operation is called with an invalid `baseModelId`. From imodel-transformer's POV, it should never be inserting elements with invalid `baseModelId` moving forward otherwise any app consuming both imodel-transformer and itjs 5.x will very likely see exception being thrown from imodel-transformer because of this illegal operation.

aruniverse:


anmolshres98:
We can probably publish and try and merge this PR

packages/transformer/src/test/IModelTransformerUtils.ts
```diff
@@ -2136,6 +2142,17 @@ export class RecordingIModelImporter extends CountingIModelImporter {
     };
     return this.targetDb.elements.insertElement(auditRecord);
   }
+  private accountForPartialViewDefinition2d(elementProps: ElementProps): void {
+    const view2d = elementProps as unknown as ViewDefinition2d;
+    // if the ViewDefinition2d has a baseModelId, it should be a valid Id64String otherwise insert/get operations will fail.
```
DanRod1999, line 2147: @khanaffan @MichaelSwigerAtBentley, would appreciate reviewing this func, as the issue started after some of you're guys work in native. I think Mike worked with Anmol a bit on this, but wanted a quick review to make sure this logic seems sound. Lgtm, but want to make sure temporarily setting this baseModelId wouldn't have unanticipated consequences before the correct value is reconciled. Thanks!

packages/performance-scripts/package.json
```diff
@@ -19,9 +19,9 @@
   "author": "Bentley Systems, Inc.",
   "license": "MIT",
   "devDependencies": {
-    "@itwin/build-tools": "^4.10.6",
-    "@itwin/core-backend": "^4.3.5",
+    "@itwin/build-tools": "^5.0.0",
+    "@itwin/core-backend": "^5.0.0",
     "@types/node": "^18.11.5",
```
MichaelSwigerAtBentley, line 24: Is it possible to bump node versions as well, since 5.x is not longer supporting node 18

packages/transformer/src/test/TestUtils/IModelTestUtils.ts
```diff
@@ -418,17 +425,38 @@ export class HubWrappers {
...
         version: IModelVersion.fromJSON(args.asOf),
         iModelId: args.iModelId,
       }),
     };
 
-    return V1CheckpointManager.getCheckpointDb({
-      checkpoint,
-      localFile: V1CheckpointManager.getFileName(checkpoint),
-    });
+    // return V2CheckpointManager.getCheckpointDb({
```
MichaelSwigerAtBentley, line 435: Any reason to keep this commented code around?

DanRod1999:
looks like updating to 5.2 broke tests, assuming tests were passing in 5.0.

Edit: these tests were mostly failing due to a change introduced in 5.1 involving changes to an elements instance key cache.

DanRod1999:
Currently 1 test is still failing
```
IModelTransformer
       transforms code values with non standard space characters:
     Error: Error inserting element [duplicate code]
      at Elements.insertElement (D:\imodel-transformer\node_modules\.pnpm\@itwin+core-backend@5.2.0_@itwin+core-bentley@5.2.0_@itwin+core-common@5.2.0_@itwin+core-bent_57r3dumqsxjt467qzcpzuya4nu\node_modules\@itwin\core-backend\src\IModelDb.ts:2316:53)
      at IModelImporter.onInsertElement (src\IModelImporter.ts:334:48)
      at IModelImporter.importElement (src\IModelImporter.ts:321:14)
      at IModelTransformer.onExportElement (src\IModelTransformer.ts:2091:21)
      at IModelExporter.exportElement (src\IModelExporter.ts:878:20)
      at IModelExporter.exportChildElements (src\IModelExporter.ts:905:9)
```         
It looks like it was also introduced between 5.1 or 5.2, but is unrelated to the fixes already made by Mike and I for other tests

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1510,20 +1513,19 @@ export class IModelTransformer extends IModelExportHandler {
     );
   }
 
-  private _queryElemIdByFedGuid(
+  // Deprecate?
+  private async _queryElemIdByFedGuid(
```
DanRod1999: converted this to async, but decided to replace current usage with itjs api `imodeldb.elements.getIdFromFederationGuid()`. This api isn't async, but I changed the apis that use this to async anyway to in case itjs ever changes. I personally don't think there's a benefit to this api anymore if the itjs one exists so I'm suggesting deprecating.

packages/transformer/src/IModelExporter.ts
```diff
@@ -745,21 +763,31 @@ export class IModelExporter {
...
-          await this._yieldManager.allowYield();
-        }
-      }
-    );
+    const params = new QueryBinder().bindId("modelId", modelId);
+    if (skipRootSubject) {
+      params.bindId("rootSubjectId", IModel.rootSubjectId);
+    }
+    for await (const row of this.sourceDb.createQueryReader(sql, params, {
+      usePrimaryConn: true,
```
DanRod1999, line 782: needed to add this because if I did not one test in catalog.test.ts would fail. Typically you can just call imodeldb.saveChanges() and that will mean you don't need to use a primary connection. However, in the failing test I could not find the correct place to call save changes on the imodeldb. I think its safe to do this since the original api always behaved this way, but wanted to point it out just in case

packages/transformer/src/IModelTransformer.ts
```diff
@@ -463,7 +460,26 @@ export class IModelTransformer extends IModelExportHandler {
         AND Identifier=:identifier
       LIMIT 1
     `;
-    // eslint-disable-next-line @itwin/no-internal, deprecation/deprecation
+
+    // if (aspectProps.scope === undefined) return undefined;
```
DanRod1999: tried updating `queryScopeExternalSourceAspect()` to use `createQueryReader()`, however this is causing the test `"should set unsafeVersions correctly when branchRelationshipDataBehavior is 'unsafe-migrate'"` to fail. My guess is it is related to this [issue](https://github.com/iTwin/itwinjs-core/issues/8942).

DanRod1999:
All usages of `withPreparedStatement()` have been replaced except for 2 instances. First occurrence is mentioned above. Second, is in the `ElementCascadingDeleteCode`. I believe I cannot update this because it is part of an override for `processElementTree()` which is in itjs core and currently sync. I don't think I can change the current code here unless the method in itjs is changed.

aruniverse:
@itwin/transformation-service can you guys please help in reviewing the changes here?

packages/test-app/src/test/Transformer.test.ts
```diff
@@ -70,17 +74,14 @@ describe("imodel-transformer", () => {
...
-          ? statement.getValue(0).getInteger()
-          : 0;
-      }
-    );
+  async function count(
+    iModelDb: IModelDb,
+    classFullName: string
+  ): Promise<number> {
+    const result = await iModelDb
+      .createQueryReader(`SELECT COUNT(*) FROM ${classFullName}`)
```
JulijaRamoskiene: Why `usePrimaryConn: true` is not used here? Same question for this query below:
```
 .createQueryReader(
    `SELECT COUNT(*) FROM ${className} e JOIN bis.Category c ON e.category.id=c.ECInstanceId WHERE c.CodeValue=:category`,
    QueryBinder.from({ category: testCategory })
)
```

packages/transformer/src/test/TestUtils/IModelTestUtils.ts
```diff
@@ -1427,10 +1463,6 @@ export class ExtensiveTestScenario {
   }
 
   public static populateDb(sourceDb: IModelDb): void {
-    // make sure Arial is in the font table
-    sourceDb.addNewFont("Arial");
-    assert.exists(sourceDb.fontMap.getFont("Arial"));
```
JulijaRamoskiene, line 1432: [Q] Why this was removed? Was this used to test how fonts are handled?

packages/transformer/src/test/standalone/IModelCloneContext.test.ts
```diff
@@ -108,6 +107,10 @@ describe("IModelCloneContext", () => {
       relationshipsProps.forEach((props) =>
         sourceDb.relationships.insertInstance(props)
       );
+
+      // Save changes to source DB to ensure relationships are persisted
+      sourceDb.saveChanges();
```
JulijaRamoskiene, line 119: [Q] Was this done to make 'createQueryReader' work? If tests need to be updated it means code is not working as before.  Is that an accepted risk? Will users be notified?

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -3271,6 +3294,8 @@ describe("IModelTransformer", () => {
       code: Code.createEmpty(),
     } as ElementProps);
 
+    sourceDb.saveChanges();
```
JulijaRamoskiene, line 3375: Again, wondering about all the additional `saveChanges` in random places, it makes me think that `withPreparedStatement` was not correctly ported to use createQueryReader with `usePrimaryConn`

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -390,26 +399,31 @@ describe("IModelTransformer", () => {
...
+        await count(
+          targetDb,
+          "ExtensiveTestScenarioTarget:TargetInformationRecord"
+        )
       );
       transformer.dispose();
     }
 
-    if (true) {
+    // removed deprecated detectElementDeletes which caused this portion to fail. Unsure if this is safe to remove so skipping for now
```
JulijaRamoskiene: I think we need to make decision now before the PR is completed. No one will ever come back to fix this part if that is not done now. I vote "Remove it now" :)

packages/transformer/src/ECReferenceTypesCache.ts
```diff
@@ -93,54 +100,85 @@ export class ECReferenceTypesCache {
...
+    const query = `
       WITH RECURSIVE refs(SchemaId) AS (
         SELECT ECInstanceId FROM ECDbMeta.ECSchemaDef WHERE Name='BisCore'
         UNION ALL
         SELECT sr.SourceECInstanceId
         FROM ECDbMeta.SchemaHasSchemaReferences sr
         JOIN refs ON sr.TargetECInstanceId = refs.SchemaId
       )
-      SELECT s.Name
+      SELECT DISTINCT s.Name as name
```
JulijaRamoskiene: `SELECT DISTINCT` is is misused here as a quick fix for poor query design. Use `Union` instead of `Union all` to remove duplicates.

packages/transformer/src/ECReferenceTypesCache.ts
```diff
@@ -93,54 +100,85 @@ export class ECReferenceTypesCache {
...
+        Logger.logTrace(
+          TransformerLoggerCategory.ECReferenceTypesCache,
+          `Completed schema: ${schemaName} in ${(endTime - startTime).toFixed(2)}ms`
+        );
+        schemaCompletedCount++;
+      } else {
+        Logger.logInfo(
+          TransformerLoggerCategory.ECReferenceTypesCache,
+          `Did not load schema: ${schemaName}`
+        );
```
JulijaRamoskiene: [Q] In what cases schema would be undefined here? Why we consider it acceptable situation and continue after logging info message?

packages/transformer/src/IModelCloneContext.ts
```diff
@@ -183,31 +183,28 @@ export class IModelCloneContext extends IModelElementCloneContext {
...
-                  sourceId: `${sourceType}${sourceId}`,
-                  targetId: `${targetType}${targetId}`,
-                } as const;
-              }
-              if (status !== DbResult.BE_SQLITE_DONE)
-                throw new IModelError(status, "unexpected query failure");
-              return undefined;
+            `;
+          const params = new QueryBinder().bindId(1, rawId);
+          const reader = this.sourceDb.createQueryReader(sql, params);
```
JulijaRamoskiene: Why usePrimaryConn: true is not used here?

packages/transformer/src/IModelCloneContext.ts
```diff
@@ -218,34 +215,36 @@ export class IModelCloneContext extends IModelElementCloneContext {
...
+          relInTargetParams.bindId(
+            1,
+            EntityReferences.toId64(relInTarget.sourceId)
+          );
+          relInTargetParams.bindId(
+            2,
+            EntityReferences.toId64(relInTarget.targetId)
           );
+          const relInTargetResult = await this.targetDb
+            .createQueryReader(relInTargetSql, relInTargetParams)
```
JulijaRamoskiene: Why usePrimaryCon: true was not used here?

packages/transformer/src/IModelCloneContext.ts
```diff
@@ -256,44 +255,50 @@ export class IModelCloneContext extends IModelElementCloneContext {
...
+      if (property.isNavigation()) {
         const sourceNavProp: RelatedElementProps | undefined =
-          sourceElementAspect.asAny[propertyName];
+          sourceElementAspect.asAny[ECJsNames.toJsName(name)];
         if (sourceNavProp?.id) {
           const navPropRefType = this._refTypesCache.getNavPropRefType(
             sourceElementAspect.schemaName,
             sourceElementAspect.className,
-            propertyName
+            ECJsNames.toJsName(name)
```
JulijaRamoskiene: [NitPick] 6 calls to convert `name` to js name, could store result to some local var instead.

packages/transformer/src/IModelExporter.ts
JulijaRamoskiene: My expectation was that all places would be ported to behave the same way as it worked before. If you needed to fix tests (call saveChanges()) it means that transformer is not behaving as before, and users that will pass iModels without calling saveChanges() will get unexpected results.

If that is the case, this should be announced to users that will update to use new version.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1258,53 +1271,41 @@ export class IModelTransformer extends IModelExportHandler {
...
-              currSourceRow.federationGuid <= currTargetRow.federationGuid)
-          ) {
-            if (sourceStmt.step() !== DbResult.BE_SQLITE_ROW) return;
-            sourceRow = sourceStmt.getRow();
-          }
-        }
-      })
-    );
+    const sourceReader = sourceDb.createQueryReader(elementIdByFedGuidQuery);
+    const targetReader = targetDb.createQueryReader(elementIdByFedGuidQuery);
```
JulijaRamoskiene: usePrimaryConn: true

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1392,56 +1389,54 @@ export class IModelTransformer extends IModelExportHandler {
...
+      FROM Bis.ExternalSourceAspect
+      WHERE Kind=?
+        AND Scope.Id=?
+        AND Identifier=?
+    `;
+    const params = new QueryBinder();
+    params.bindString(1, ExternalSourceAspect.Kind.Relationship);
+    params.bindId(2, this.targetScopeElementId);
+    params.bindString(3, entityInProvenanceSourceId);
+    const result = (await this.provenanceDb()).createQueryReader(sql, params);
```
JulijaRamoskiene: usePrimaryConn: true?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1451,79 +1446,75 @@ export class IModelTransformer extends IModelExportHandler {
...
+        and ecclassid=?
+    `;
+    const params = new QueryBinder();
+    params.bindId(1, targetRelInfo.sourceId);
+    params.bindId(2, targetRelInfo.targetId);
+    params.bindId(
+      3,
+      await this._targetClassNameToClassId(sourceRelInfo.classFullName)
     );
+    const result = this.targetDb.createQueryReader(sql, params);
```
JulijaRamoskiene: Why not `usePrimaryConn: true`?

packages/transformer/src/IModelTransformer.ts
JulijaRamoskiene: Just remove it as it is private .

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1539,53 +1530,54 @@ export class IModelTransformer extends IModelExportHandler {
...
-        }
-        const targetElemId = stmt.getValue(1).getId();
-        const wasDeletedInSource = !EntityUnifier.exists(this.sourceDb, {
-          entityReference: `e${aspectIdentifier}`,
-        });
-        if (wasDeletedInSource) this.importer.deleteElement(targetElemId);
-      }
-    });
-  }
+  // public async detectElementDeletes(): Promise<void> {
```
JulijaRamoskiene: why this was commented out instead of removed? we have git history if we want to revert this.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2116,7 +2110,14 @@ export class IModelTransformer extends IModelExportHandler {
     if (!this._options.wasSourceIModelCopiedToTarget) {
       this.importer.importElement(targetElementProps); // don't need to import if iModel was copied
     }
-    this.context.remapElement(sourceElement.id, targetElementProps.id!); // targetElementProps.id assigned by importElement
+
+    if (targetElementProps.id === undefined) {
+      throw new IModelError(
+        IModelStatus.BadElement,
+        "targetElementProps.id should be assigned by importElement"
```
JulijaRamoskiene, line 1301: [NitPick] Misleading message. ImportElement is not called when `wasSourceIModelCopiedToTarget` is true. Suggest to change it to something like `targetElementProps.id should be already assigned.`

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2215,31 +2218,12 @@ export class IModelTransformer extends IModelExportHandler {
...
-              return true;
-            case DbResult.BE_SQLITE_DONE:
-              return false;
-            default:
-              assert(false, `unexpected db result: '${JSON.stringify(stmt)}'`);
-          }
-        }
-      );
+      const params = new QueryBinder().bindId("targetModelId", targetModelId);
+      const reader = this.targetDb.createQueryReader(sql, params);
```
JulijaRamoskiene: Why not usePrimaryConn:true?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2667,58 +2652,52 @@ export class IModelTransformer extends IModelExportHandler {
...
-            }
-            aspectDeleteIds.push(statement.getValue(0).getId());
-          }
-          await this._yieldManager.allowYield();
-        }
-      }
-    );
-    this.targetDb.elements.deleteAspect(aspectDeleteIds);
-  }
+  // public async detectRelationshipDeletes(): Promise<void> {
```
JulijaRamoskiene: Remove commented out code

packages/transformer/src/IModelTransformer.ts
```diff
@@ -3251,30 +3227,41 @@ export class IModelTransformer extends IModelExportHandler {
...
       // since we are processing one changeset at a time, we can see local source deletes
       // of entities that were never synced and can be safely ignored
       const deletionNotInTarget = !targetId;
       if (deletionNotInTarget) return;
-      this.context.remapElement(changedInstanceId, targetId!);
+
+      if (targetId === undefined) {
+        throw new IModelError(
+          IModelStatus.BadElement,
+          "targetId should aquired from source id or element provenance"
```
JulijaRamoskiene: fix spelling

packages/transformer/src/IModelTransformer.ts
```diff
@@ -3431,27 +3412,29 @@ export class IModelTransformer extends IModelExportHandler {
...
     ) {
-      // eslint-disable-next-line deprecation/deprecation
-      await this.detectElementDeletes();
-      // eslint-disable-next-line deprecation/deprecation
-      await this.detectRelationshipDeletes();
+      Logger.logWarning(
+        loggerCategory,
+        "This workflows was deprecated in v1 and is no longer supported"
+      );
+      //   await this.detectElementDeletes();
```
JulijaRamoskiene: commented out code

packages/transformer/src/IModelTransformer.ts
```diff
@@ -3672,25 +3661,24 @@ export class TemplateModelCloner extends IModelTransformer {
...
     return targetElementProps;
   }
 }
 
-function queryElemFedGuid(db: IModelDb, elemId: Id64String) {
-  // eslint-disable-next-line @itwin/no-internal, deprecation/deprecation
-  return db.withPreparedStatement(
-    `
+//Deprecate in preference of imodeldb.elements.getFederationGuidFromId()?
+async function queryElemFedGuid(
```
JulijaRamoskiene: why deprecate instead of remove?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -793,15 +809,19 @@ export class IModelTransformer extends IModelExportHandler {
   /** Return the IModelDb where IModelTransformer will store its provenance.
    * @note This will be [[targetDb]] except when it is a reverse synchronization. In that case it be [[sourceDb]].
    */
-  public get provenanceDb(): IModelDb {
-    return this.isReverseSynchronization ? this.sourceDb : this.targetDb;
+  public async provenanceDb(): Promise<IModelDb> {
```
JulijaRamoskiene: suggest to follow same naming guidelines with all refactored getters (same aas `getIsForwardSynchronization`). Same comment for: provenanceSourceDb and synchronizationVersion

packages/transformer/src/test/TestUtils/IModelTestUtils.ts
DanRod1999: these look like they are probably also withing itjs iModetestUtils as I think they were copied from there. I currently have no real plans updating this package except for potentially copying over cores version into here

packages/transformer/src/IModelTransformer.ts
DanRod1999: I was unsure if we are allowed to just remove without first deprecating. I know in itjs this would be problematic. Doubt its an issue here, but not sure if we should be following any guidelines

packages/transformer/src/test/TestUtils/IModelTestUtils.ts
DanRod1999: these apis were changed in 5.x. `addNewFont()` no longer exists and `fontmap` was deprecated. However, there is an over arching problem with this file in general because its basically a trimmed down copy of IModelTestUtils.ts in itwinjs-core repo. I don't fully understand why me have to copy these utils over. I assume the one in itwinjs-core is just private, and someone just decided to to this instead.

My true intention for this file is to update it with the version in itwinjs-core, but I personally don't want to do that in this PR. Partially because its would add clutter to this file that doesn't actually warrant a review, and also because itwinjs-core still has usage of deprecated functions so it seems weird to update now.

DanRod1999: that being said, i didn't realize how many changes already went into this file, so maybe I should just update now? Not sure, open to opinions

packages/transformer/src/test/standalone/IModelCloneContext.test.ts
DanRod1999: After discussing this, we think its best to set `{usePrimaryConn: true}` however I'm going to keep these save changes I think since I don't think they are bad practice? Unless you think its better to revert to prove functionality is unchanged?

packages/transformer/src/IModelTransformer.ts
aruniverse: we should query for usage of the api across gh and ado

DanRod1999: It really shouldn't exist outside of transformer because itjs has its own

packages/transformer/src/test/TestUtils/IModelTestUtils.ts
DanRod1999: just going to update this later for now

packages/transformer/src/test/IModelTransformerUtils.ts
```diff
@@ -512,28 +511,33 @@ const aliasedProperties: Record<string, Record<string, string> | undefined> =
...
+      const base = elem.iModel.schemaContext.getSchemaItemSync(
+        property.class.fullName
       );
-    }
+      if (base !== undefined && base instanceof Entity) {
+        Object.assign(allProperties, getAllClassMetaDataProperties(name, base));
+      }
+    });
 
     Object.assign(allProperties, aliasedProperties[className.toLowerCase()]);
```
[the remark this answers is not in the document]
DanRod1999: I don't feel that this comment is necessary as it seems like the functions that were used to replace this are correct, and the prior impl was incorrect?

packages/transformer/src/ECReferenceTypesCache.ts
DanRod1999: This was part of a change from copilot that was made to get some performance improvement from an issue in Substation+. I didn't have a chance to review this myself, so I'm doing that now, and will clean up this stuff. I believe based on the original impl this expected to always locate the schema, so maybe throwing an error would be more appropriate

DanRod1999: reworked this, so now it will error if that occurs, however, not sure if that would ever happen

packages/transformer/src/EntityUnifier.ts
```diff
@@ -65,21 +64,15 @@ export namespace EntityUnifier {
...
-      `SELECT 1 FROM ${classFullName} WHERE ECInstanceId=?`,
-      (stmt) => {
-        stmt.bindId(1, id);
-        const matchesResult = stmt.step();
-        if (matchesResult === DbResult.BE_SQLITE_ROW) return true;
-        if (matchesResult === DbResult.BE_SQLITE_DONE) return false;
-        else throw new IModelError(matchesResult, "query failed");
-      }
-    );
+    const sql = `SELECT 1 FROM ${classFullName} WHERE ECInstanceId=?`;
```
DanRod1999: This appears to be causing a major perf regression. If returned back to withPreparedStatement() perf is better, but it seems like createQueryReader() should not be used for these smaller more frequent queries? Not sure if there is any solution for this atm except for creating a cache? `withPreparedSqlStatement()` might also be an option?

MichaelSwigerAtBentley: Have you asked @khanaffan about this? Seems crazy that swapping to createQueryReader caused such a perf drop.

DanRod1999: not yet, but would value his input. My initial thoughts were I know `withPreparedStatement()` was particularly useful for repetitive queries compared to `withStatement()` and maybe `createQueryReader()` behaved similarly?

DanRod1999: Here's an example of some results transforming an imodel with 10k elements: 
IModelTransformer Performance Tests

Results: createQueryReader()
-  Elements: 10000
-  Element insertion: 1602.74 ms
-  Schema processing: 6.57 ms
-  process() duration: 10350.41 ms
-  Avg per element: 1.0350 ms

Results: withPreparedStatement():
-  Elements: 10000
-  Element insertion: 1847.14 ms
-  Schema processing: 6.69 ms
-  process() duration: 6659.28 ms
-  Avg per element: 0.6659 ms

MichaelSwigerAtBentley: Yeah, I would assume we need an async alternative with the repetitive query performance of withPreparedStatement() - Is there an optimization of the query that could be done? 

CC: @RohitPtnkr1996 might be able to assist too

khanaffan: We have that in a PR https://github.com/iTwin/itwinjs-core/pull/8970

CC: @soham-bentley

DanRod1999: > We have that in a PR https://github.com/iTwin/itwinjs-core/pull/8970

but to clarify, this PR is for synchronous operations. We do not currently have a solution for async operations in this case. For me this would be fine as I don't require async, but I guess my questions are: do we need a async solution for this slowdown, and did we know this could potentially cause such a dramatic perf hit?

DanRod1999:
@copilot resolve the merge conflicts in this pull request

packages/performance-tests/test/iModelUtils.ts
```diff
@@ -31,16 +30,7 @@ export interface IModelParams {
 // TODO: dedup with other packages
 // for testing purposes only, based on SetToStandalone.ts, force a snapshot to mimic a standalone iModel
 export function setToStandalone(iModelPath: string) {
```
JulijaRamoskiene: [NitPick] TODO no longer relevant

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -390,80 +399,25 @@ describe("IModelTransformer", () => {
...
+        await count(sourceDb, "ExtensiveTestScenario:SourceInformationRecord")
       );
       assert.equal(
         3,
-        count(targetDb, "ExtensiveTestScenarioTarget:TargetInformationRecord")
-      );
-      transformer.dispose();
-    }
-
-    if (true) {
```
JulijaRamoskiene, line 412: Q: Why this was deleted? Want to make sure that some bug is not introduced due to which test started failing

packages/transformer/src/test/TestUtils/IModelTestUtils.ts
```diff
@@ -1427,10 +1463,6 @@ export class ExtensiveTestScenario {
   }
 
   public static populateDb(sourceDb: IModelDb): void {
-    // make sure Arial is in the font table
```
JulijaRamoskiene, line 1430: Q: Why font registration and default view configuration was removed? Want to make sure that some bug is not introduced due to which test started failing.

There was a fix for project extent, due to which some views were not parsed, maybe issue is now fixed?

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -4365,190 +4324,6 @@ describe("IModelTransformer", () => {
     targetDb.close();
   });
 
-  it("detect element deletes works on children", async () => {
```
JulijaRamoskiene, line 4368: [Q] Why tests for deletes were removed?

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -556,28 +516,27 @@ describe("IModelTransformer", () => {
...
-          ? statement.getValue(0).getInteger()
-          : 0;
-      }
-    );
+  async function count(
+    iModelDb: IModelDb,
+    classFullName: string
+  ): Promise<number> {
+    const sql = `SELECT COUNT(*) FROM ${classFullName}`;
+    const reader = iModelDb.createQueryReader(sql);
```
JulijaRamoskiene: Use primary conn here, not all tests call saveChanges()

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -399,13 +420,20 @@ describe("IModelTransformerHub", () => {
...
+          {}
+        );
         await transformer.process();
+        // // Verify processAll wrote the sync version so subsequent processChanges starts from correct index
+        // const syncVersionAfterProcessAll = await transformer["_provenanceManager"].getSynchronizationVersion();
+        // assert.equal(
+        //   syncVersionAfterProcessAll.index,
+        //   sourceDb.changeset.index,
+        //   "processAll should write sync version matching source changeset index"
+        // );
```
JulijaRamoskiene: Did you mean to uncomment this?

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -3535,6 +3571,82 @@ describe("IModelTransformerHub", () => {
     sinon.restore();
   });
 
+  // Regression test for https://github.com/iTwin/imodel-transformer/issues/28
+  it("should succeed when element is deleted and element with the same code is re-added in the next changeset", async () => {
```
JulijaRamoskiene, line 3741: Thanks 1000 times for this :)

packages/transformer/src/IModelExporter.ts
```diff
@@ -629,19 +628,51 @@ export class IModelExporter {
    */
   public async exportFonts(): Promise<void> {
     Logger.logTrace(loggerCategory, "exportFonts()");
-    for (const font of this.sourceDb.fontMap.fonts.values()) {
-      await this.exportFontByNumber(font.id);
+    for (const font of this.sourceDb.fonts.queryMappedFamilies()) {
+      await this.exportFontByFontProps(font);
```
JulijaRamoskiene, line 632: [Q] Why `this.exportFontByFontProps` is called instead of ` await this.exportFontByNumber(font.id);` now?

packages/transformer/src/ECReferenceTypesCache.ts
```diff
@@ -93,54 +100,80 @@ export class ECReferenceTypesCache {
...
     }
+
+    this._rootBisClassCache.set(cacheKey, bisRootForConstraint);
     return bisRootForConstraint;
   }
 
   private async getAbstractConstraintClass(
     constraint: RelationshipConstraint
   ): Promise<ECClass> {
+    const cacheKey = `${constraint.fullName}_${constraint.constraintClasses?.[0]?.fullName || "abstract"}`;
```
JulijaRamoskiene: Can same cache key be generated for 2 different constraints that have same first constraint class?

packages/transformer/src/IModelExporter.ts
DanRod1999: I'll double check this change but I believe it was due to deprecated changes with some font related apis

DanRod1999: `this.sourceDb.fontMap` is deprecated in 5.x, so I switched to `fonts.queryMappedFamilies()` and that returns a font prop object, so no need to export by the font id. `exportFontByFontProps()` is more direct way to export font.

Call Stack:
`this.exportFontByNumber(fontId)` ->
`this.exportFontByFontFamilyDescriptor(fontFamilyDescriptor)` ->
`this.exportFontByFontProps({ ...fontFamily, id: fontId });`

JulijaRamoskiene (changes requested):

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
DanRod1999: yeah not sure why I commented it out, but I reverted

packages/transformer/src/test/standalone/IModelTransformer.test.ts
DanRod1999: good catch thanks

packages/transformer/src/test/standalone/IModelTransformer.test.ts
DanRod1999: this portion failed after detectElementDeletes was removed. At some point i had a comment about this.

DanRod1999: https://github.com/iTwin/imodel-transformer/pull/234#discussion_r2798709243
Related to comment left here.

Since detectElementDeletes was removed, this workflow I think is not longer appropriate

packages/transformer/src/test/standalone/IModelTransformer.test.ts
DanRod1999: detectElementDeletes which was deprecated is now removed, so I got rid of the tests related to them

packages/transformer/src/test/TestUtils/IModelTestUtils.ts
DanRod1999: deleted cause I thought it was irrelevant for our uses and the original api was deprecated. Found the correct upgrade path for that api and am using that instead.

In general these apis will need to be updated to match the IModelTestUtils.ts in itwinjs-core. I can't port over the new changes yet because the updated functions now use edit txns, but I will update accordingly, when I resolve those deprecation errors.

packages/transformer/src/ECReferenceTypesCache.ts
DanRod1999: I don't believe so, but I updated it a bit to assure uniqueness. Now it includes the schema, relationship name, and source or target. Whereas before it was only the relationship name, so if 2 schemas had the same relationship name their could have been a collision.

PR #235, opened 2025-02-11
Getting started with imodel-transformer

derbynn:
- documentation to get started


docs/learning/transformer/getting-started.md
```diff
@@ -0,0 +1,87 @@
+# Getting Started with iModel-Transformer
+
+## Overview
+
+The repository supports both JavaScript and TypeScript. To build a strong foundation, check out these resources:
```
ben-polinsky, line 5: Is this section on TS/JS necessary?

docs/learning/transformer/getting-started.md
```diff
@@ -0,0 +1,87 @@
...
+  [The Modern JavaScript Tutorial](https://javascript.info/)  
+  [Mozilla JavaScript Reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference)
+  
+- **TypeScript:**  
+  [TypeScript Documentation](https://www.typescriptlang.org/docs/)
+
+For more details on the transformer itself:
+
+- **Transformer Repo:** [GitHub - imodel-transformer](https://github.com/iTwin/imodel-transformer)
+- **Transformer Docs:** [iTwinJS Transformer Learning](https://www.itwinjs.org/learning/transformer)
```
ben-polinsky, line 17: Presumably this document is going into the docs - I don't think that a link to "Transformer Docs" makes sense.

docs/learning/transformer/getting-started.md
```diff
@@ -0,0 +1,87 @@
...
+- **Branching & Synchronization:** [Branching iModels Documentation](https://github.com/iTwin/imodel-transformer/blob/main/docs/learning/transformer/branching-imodels.md)
```
ben-polinsky, line 18: This should be a relative link since the target exists within the trasnformer docs. You can use `$docs` to represent the root of the doc site (i.e. `$docs/learning/trasnformer/branching-imodels.md`).

docs/learning/transformer/getting-started.md
```diff
@@ -0,0 +1,87 @@
...
+
+- **Checkpoints:**  
+  A checkpoint is a full snapshot of the iModel at a given point, with changesets already applied. This can be used to reduce number of Changesets needed to apply to get to a certain version of the iModel.
+  - **Checkpoint X** might represent the state after changesets 1–3, letting you quickly access **Point A**.
+  - To reach **Point B**, you might download **Checkpoint Y** and apply only the remaining changes (e.g., changeset 6).
+
+## Key Concepts
+
+- **Snapshot:**  
+  A [static](https://www.itwinjs.org/learning/backend/accessingimodels/#snapshot-imodels) point-in-time representation of an iModel’s state.
```
ben-polinsky, line 43: See comment above about relative links and using $docs. There are several more instances in this document as well.

docs/learning/transformer/getting-started.md
ben-polinsky: Understood, but if this is going in the transformer package docs, the audience is much larger.

docs/learning/transformer/getting-started.md
derbynn: It seems to me that $docs is not an in-built thing in markdown file? I see that there are some other broken links is why there are also broken links in the existing transformer documentation - https://github.com/iTwin/imodel-transformer/blob/main/docs/learning/transformer/branching-imodels.md. Do the $docs tag only work when this markdown file is published to itwin.js org docs?

PR #238, opened 2025-02-21
Ignore navigational properties that point to an ECView

ViliusRuskys:
ECView elements are not "real" data and transformer does not need to process them. Relationships that point to an ECView instance are only allowed on ECView elements themselves, hence we don't need to track these relationships.

### DIFF
--- a/packages/transformer/src/test/standalone/ECReferenceTypesCache.test.ts
+++ b/packages/transformer/src/test/standalone/ECReferenceTypesCache.test.ts
@@ -13,12 +13,17 @@ import { KnownTestLocations as BackendTestsKnownLocations } from "../TestUtils/K
 import * as Semver from "semver";
 import { Schema, SchemaItemType, SchemaLoader } from "@itwin/ecschema-metadata";
 import * as sinon from "sinon";
+import { version as iTwinCoreBackendVersion } from "@itwin/core-backend/package.json";
 
 describe("ECReferenceTypesCache", () => {
   let testIModel: SnapshotDb;
   const testSchemaPath = path.join(
     BackendTestsKnownLocations.assetsDir,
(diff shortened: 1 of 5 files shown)


mindaugasdirg (dismissed):
Looks good to me

ViliusRuskys:
@aruniverse Does anyone from your team want to look at this PR, or should I just merge it?
I have 1 approval from the transformation service's team side, but we are not the main contributors to this project so I'm hesitant to merge it.

packages/transformer/src/test/standalone/ECReferenceTypesCache.test.ts
```diff
@@ -185,6 +190,20 @@ describe("ECReferenceTypesCache", () => {
...
+      return; // Pre 4.6.0 does not have QueryView support. https://www.itwinjs.org/bis/domains/ecdbmap.ecschema/#queryview
+    }
+    const thisTestRefCache = new ECReferenceTypesCache();
+    const ecdbMapVersion =
+      emptyWithBrandNewBiscore.querySchemaVersion("ECdbMap");
+    assert(ecdbMapVersion !== undefined);
+    assert(Semver.gte(ecdbMapVersion, "2.0.4"));
+    await emptyWithBrandNewBiscore.importSchemas([testSchemaPathWithQueryView]);
+    emptyWithBrandNewBiscore.saveChanges();
+    await thisTestRefCache.initAllSchemasInIModel(emptyWithBrandNewBiscore);
```
DanRod1999, line 204: How do we know if this test were to fail? Do we just expect initAllSchemasInIModel() to always produce an error if it failed? Is there anything on thisTestRefCache that we can assert to prove the changes from relInfoFromRelClass worked?

Issue #231, opened 2025-02-04
Line style information is not exported


ViliusRuskys:
Custom line styles are not exported to the target iModel and are lost.

Custom LineStyle information is not stored on the LineStyle element. The element actually stores a reference in its `data` property to a row in the be_Prop table. This be_Prop row is the one that stores the actual geometry for the line style.

I posted a discussion thread on the be_Prop table. One of replies pointed out that there is some WIP code on native side that might be helpful. 
https://github.com/iTwin/itwinjs-core/discussions/7642

DanRod1999:
Do you have a test I can use to repro where lines styles should be exported to a target iModel. I tried tracing the code myself from the transformer repo when calling `exportModel`, but only saw the `getElement()` function being called, and I see that it takes a "wantGeometry" which I believe is important. Brien suggested using some of the methods from `GeometryStreamIO::Import()`, however, it appears this import function is not currently a part of getElement.

So far I only checked `getElement()` because I don't know where else we might be hitting the native code in the transformer. Is the `exportModel` function where you are seeing this issue, or is it something else? Ideally I would like to get a call stack in the native code using whatever functions you are calling, and see if we call the functions Brien is referring to. If not we may need to try find a way to call the `_RemapLineStyleId` function he pointed out.

ViliusRuskys:
I've created a simple iModel with a couple of lines where this issue is reproducible.
You can use this code to create the source iModel:
```
const subjectId = Subject.insert(iModelDb, IModel.rootSubjectId, "My objects");
const defModelId = DefinitionModel.insert(iModelDb, subjectId, "DefinitionModel");

// Entires in be_Prop table:
const lsStrokes1: LineStyleDefinition.Strokes = [];
const lsStrokes2: LineStyleDefinition.Strokes = [];
lsStrokes1.push({ length: 4, strokeMode: LineStyleDefinition.StrokeMode.Dash });
lsStrokes1.push({ length: 2 });
lsStrokes2.push({ length: 1, strokeMode: LineStyleDefinition.StrokeMode.Dash });
lsStrokes2.push({ length: 1 });
const styleProps1 = LineStyleDefinition.Utils.createStrokePatternComponent(iModelDb, { descr: "line1", strokes: lsStrokes1 });
const styleProps2 = LineStyleDefinition.Utils.createStrokePatternComponent(iModelDb, { descr: "line2", strokes: lsStrokes2 });

// LineStyle Elements
const style1: LineStyleProps = {
[...]

PR #242, opened 2025-03-10
Tests for Importing Lines Styles

DanRod1999:
Test for importing regular and compound line styles when preforming process all on imodels with line style elements. Tested on changes from:
https://github.com/iTwin/imodel-native/pull/1024

Original issue:
https://github.com/iTwin/imodel-transformer/issues/231

This pr is targeted for itwinjs-v5 as when I was working on the changes in imodel-native it was easier to work on imodel-native main vs release branches. Should these changes be needed in 4.x we can back port the native changes and these tests


DanRod1999:
Closing since this branch is out of date, and it would be simpler to just rewrite tests in new PR. I'm also still not convinced this is the best way to test this functionality as the fix to this was all in imodel-native, and a proper test would be in itwinjs-core. AI might be able to help write a better set of tests now in core.

PR #249, opened 2025-04-03
Update itwin versions and add tests for line styles

DanRod1999:
Add tests for importing line styles, and bump itwin versions to consume new core-backend package with updated imodel-native code that fixes the bug with imported line styles.


packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -5260,4 +5272,356 @@ describe("IModelTransformer", () => {
...
+          assert.equal(row.name, "LineCodeV1");
+          rowCount++;
+        }
+        assert.equal(rowCount, 2);
+      }
+    );
+
+    transformer.dispose();
+    iModelDb.close();
+    targetDb.close();
```
DanRod1999, line 5428: need to delete db's after test

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -3498,7 +3498,7 @@ describe("IModelTransformerHub", () => {
         master: {
           manualUpdate(db) {
             const notDeleted = db.elements.deleteDefinitionElements([
-              spatialViewDef.id,
+              spatialViewDef!.id,
```
DanRod1999, line 3501: Not sure what happened, but when updating packages some how this broke, and "!" was needed. This test is skipped, so I don't know about its importance, but I am concerned I updated something I should not have

packages/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -5260,4 +5272,356 @@ describe("IModelTransformer", () => {
...
+  });
+
+  it.only("should import compound line style from geometry stream", async function () {
```
DanRod1999, line 5431: remove only

ViliusRuskys (dismissed):

MichaelSwigerAtBentley (dismissed):

ViliusRuskys (approved):

MichaelSwigerAtBentley (approved):

PR #258, opened 2025-04-25
Fix for exporting custom changes when iModel has no changesets.

### LINKED ISSUE #257: Exporter ignores custom changes when source has no changesets pushed
**Describe the bug**
Process changes workflow reverts to export all functionality when source has no change sets. This results in custom changes being ignored.

**Expected behavior**
Custom changes should be applied when source has no change sets.

JulijaRamoskiene:
Fixed exporting custom changes when source has no change sets.
Fixes #257

Added a separate issue to discuss if this fall back is at all necessary #259

### DIFF
--- a/packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
+++ b/packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
@@ -4915,9 +4915,6 @@ describe("IModelTransformerHub", () => {
       public override async addCustomChanges(
         _sourceDbChanges: ChangedInstanceIds
       ) {}
-      public override shouldExportElement(sourceElement: Element) {
-        return super.shouldExportElement(sourceElement);
-      }
     }
 
     it("should call addCustomChanges when processing changes after source and target id map is populated", async () => {
@@ -4998,12 +4995,7 @@ describe("IModelTransformerHub", () => {
 
       // === Transformation 1: Run `process all` transformation ===
       let transformer = new CustomChangesTransformer(sourceDb, targetDb, false);
-      sinon
-        .stub(transformer, "shouldExportElement")
-        .callsFake((sourceElement) => {
-          // Exclude all drawings
-          return sourceElement.id !== documentListModel;
-        });
+      transformer.exporter.excludeElement(documentListModel);
       await transformer.process();
       transformer.updateSynchronizationVersion({
         initializeReverseSyncVersion: true,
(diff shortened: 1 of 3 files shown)


packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -5809,6 +5809,87 @@ describe("IModelTransformerHub", () => {
...
+      sourceDb.saveChanges("Initial changes");
+
+      // process all
+      const transformer1 = new CustomChangesTransformer(
+        sourceDb,
+        targetDb,
+        false
+      );
+      sinon
+        .stub(transformer1, "shouldExportElement")
```
ViliusRuskys, line 5843: Why are we stubbing this? Why not just call `transformer1.exporter.ExcludeElement(<id>)`?

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -5809,6 +5809,87 @@ describe("IModelTransformerHub", () => {
...
+      expect(targetDb.elements.tryGetElement(subjectFedGuid2)).to.be.undefined;
+
+      // process changes
+      const transformer2 = new CustomChangesTransformer(
+        sourceDb,
+        targetDb,
+        true
+      );
+      const addChangesStub = sinon
+        .stub(transformer2, "addCustomChanges")
```
ViliusRuskys, line 5864: Why is stubbing needed? Can we just add custom changes normally?

JulijaRamoskiene: I would need to implement "normal" way in `CustomChangesTransformer` class. Each test wants different action to be done there e.g, I verify if elements are remapped in some tests.

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
ViliusRuskys: If `shouldExportElement` implementation changes in the future that breaks this test, this test will not fail. I'd rather keep the entire transformer original because that is the object that we are testing.
Also, excluding the element in the intended way is simpler and less code 😄

JulijaRamoskiene: Replaced ``shouldExportElement` stubs with calls to `excludeElement`.

ViliusRuskys (approved):

DanRod1999 (approved):

PR #260, opened 2025-05-28
Transform Linearly Located Imodels

DanRod1999:
Creates transform that can be applied to spatial elements when cloning a src imodel into target with different ECEF locations.

### DIFF
--- a/packages/transformer/src/IModelExporter.ts
+++ b/packages/transformer/src/IModelExporter.ts
@@ -293,6 +293,8 @@ export class IModelExporter {
   private _progressCounter: number = 0;
   /** Optionally cached entity change information */
   private _sourceDbChanges?: ChangedInstanceIds;
+  /** A flag that determines if spatial elements should be transformed if the src and target ECEF's of the imodels differ */
+  public linearlyTransformSpatialElements: boolean = false;
 
   /**
    * Retrieve the cached entity change information.
--- a/packages/transformer/src/IModelExporter.ts
+++ b/packages/transformer/src/IModelExporter.ts
@@ -293,10 +293,6 @@ export class IModelExporter {
   private _progressCounter: number = 0;
   /** Optionally cached entity change information */
   private _sourceDbChanges?: ChangedInstanceIds;
-  /** A flag that determines if spatial elements in the source db should be transformed if source and target ECEF's of the imodels differ
-   * @note This flag should only be used if imodels are linearly located
-   */
-  public alignECEFLocations: boolean = false;
 
   /**
    * Retrieve the cached entity change information.
--- a/packages/transformer/src/IModelExporter.ts
+++ b/packages/transformer/src/IModelExporter.ts
@@ -293,8 +293,10 @@ export class IModelExporter {
   private _progressCounter: number = 0;
   /** Optionally cached entity change information */
   private _sourceDbChanges?: ChangedInstanceIds;
-  /** A flag that determines if spatial elements should be transformed if the src and target ECEF's of the imodels differ */
-  public linearlyTransformSpatialElements: boolean = false;
+  /** A flag that determines if spatial elements in the source db should be transformed if source and target ECEF's of the imodels differ
+   * @note This flag should only be used if imodels are linearly located
+   */
+  public alignECEFLocations: boolean = false;
 
   /**
    * Retrieve the cached entity change information.
--- a/packages/transformer/src/IModelTransformer.ts
+++ b/packages/transformer/src/IModelTransformer.ts
@@ -380,6 +380,10 @@ export class IModelTransformer extends IModelExportHandler {
   public readonly targetDb: IModelDb;
   /** The IModelTransformContext for this IModelTransformer. */
   public readonly context: IModelCloneContext;
+  /** The transform to be applied to the placement of spatial elements when source and target db have different ECEF locations
+   * @note transform can only be used when source and target are linearly located imodels
+   */
+  public ecefTransform?: Transform;
   private _syncType?: SyncType;
 
   /** The Id of the Element in the **target** iModel that represents the **source** repository as a whole and scopes its [ExternalSourceAspect]($backend) instances. */
@@ -1574,7 +1578,7 @@ export class IModelTransformer extends IModelExportHandler {
     }
 
     if (
-      this.exporter.linearlyTransformSpatialElements === true &&
+      this.ecefTransform !== undefined &&
       sourceElement instanceof GeometricElement3d
     ) {
       // can check the sourceElement since this IModelTransformer does not remap classes
@@ -1582,9 +1586,8 @@ export class IModelTransformer extends IModelExportHandler {
         (targetElementProps as GeometricElement3dProps).placement
       );
 
-      const ecefTransform = this.getEcefTransform(this.sourceDb, this.targetDb);
       if (placement.isValid) {
-        placement.multiplyTransform(ecefTransform);
+        placement.multiplyTransform(this.ecefTransform);
         (targetElementProps as GeometricElement3dProps).placement = placement;
       }
     }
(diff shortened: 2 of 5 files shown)


packages/transformer/src/IModelExporter.ts
```diff
@@ -293,6 +293,8 @@ export class IModelExporter {
   private _progressCounter: number = 0;
   /** Optionally cached entity change information */
   private _sourceDbChanges?: ChangedInstanceIds;
+  /** A flag that determines if spatial elements should be transformed if the src and target ECEF's of the imodels differ */
+  public linearlyTransformSpatialElements: boolean = false;
```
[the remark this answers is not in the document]
ben-polinsky: Should the name convey what you write in the comment, that the transform is because the ECEFs differ?
`alignECEFLocations`?

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1571,9 +1572,65 @@ export class IModelTransformer extends IModelExportHandler {
         targetElementProps.jsonProperties.Subject.Job = undefined;
       }
     }
+
+    if (
+      this.exporter.linearlyTransformSpatialElements === true &&
```
ben-polinsky, line 1577: won't `this.exporter.linearlyTransformSpatialElements` suffice?

packages/transformer/src/test/standalone/GeoLocationTransform.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    dbName: string,
+    numElements: number = 1,
+    color: string = "red"
+  ): SnapshotDb {
+    const dbFileName = initOutputFile(`${dbName}.bim`);
+    const imodelDb = SnapshotDb.createEmpty(dbFileName, {
+      rootSubject: { name: dbName },
+      ecefLocation: ecef,
+    });
+    imodelDb.setEcefLocation(ecef);
```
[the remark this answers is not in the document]
tcobbs-bentley: https://github.com/iTwin/itwinjs-core/blob/5b3fd1f0e410d9ea677b66b4004b8ef13e53d3e8/core/backend/src/IModelDb.ts#L3519-L3520

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1571,9 +1572,65 @@ export class IModelTransformer extends IModelExportHandler {
...
+      sourceElement instanceof GeometricElement3d
+    ) {
+      // can check the sourceElement since this IModelTransformer does not remap classes
+      const placement = Placement3d.fromJSON(
+        (targetElementProps as GeometricElement3dProps).placement
+      );
+
+      const ecefTransform = this.getEcefTransform(this.sourceDb, this.targetDb);
```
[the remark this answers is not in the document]
DanRod1999: I corrected this by creating a new optional attribute in the IModelTransformer class and setting the attribute during `process()`. Not sure if this is the best/correct solution, but it does resolve this redundant calculation. I don't want to add unnecessary attributes, but can't think of a more appropriate solution.

ben-polinsky (approved):

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
DanRod1999: Ah I missed the `IModelTransformOptions` class. Probably better to move my alignEcefLocation there instead of in the exporter? I'll see what that looks like, but atm that seems much more appropriate.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1571,9 +1585,69 @@ export class IModelTransformer extends IModelExportHandler {
...
   }
 
+  /**
+   * Calculate the transform between two ECEF locations
+   * @param srcDb
+   * @param targetDb
+   * @returns Transform that converts relative coordinates in the source iModel to relative coordinates in the target iModel.
+   * @note This can only be used if both source and target iModels are linearly located
+   */
+  public getEcefTransform(srcDb: IModelDb, targetDb: IModelDb): Transform {
```
ben-polinsky, line 1613: nit: the `get` prefix keeps bothering me because it feels like you're getting something from the Transformer, rather than computing something from two iModels... `calculateEcefTransform`? Please disregard if you hate this.

ben-polinsky (dismissed):

ben-polinsky (approved):

ben-polinsky (dismissed):

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1571,9 +1585,72 @@ export class IModelTransformer extends IModelExportHandler {
...
       }
     }
+
+    if (
+      this.ecefTransform !== undefined &&
+      sourceElement instanceof GeometricElement3d
+    ) {
+      // can check the sourceElement since this IModelTransformer does not remap classes
+      const placement = Placement3d.fromJSON(
+        (targetElementProps as GeometricElement3dProps).placement
```
mindaugasdirg, line 1595: Maybe it's worth to check that targetElementProps are `geometricElement3dProps` in case someone decides to use `this.context.remapClass` to remap geometric element to something else.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1571,9 +1585,72 @@ export class IModelTransformer extends IModelExportHandler {
...
+   */
+  public calculateEcefTransform(
+    srcDb: IModelDb,
+    targetDb: IModelDb
+  ): Transform {
+    const srcEcefLoc = srcDb.ecefLocation;
+    const targetEcefLoc = targetDb.ecefLocation;
+
+    if (
+      srcDb.geographicCoordinateSystem !== undefined &&
```
mindaugasdirg, line 1621: Probably or (`||`) should be used here since check won't pass if one of the coordinate systems is defined

mindaugasdirg (changes requested):

packages/transformer/src/IModelTransformer.ts
DanRod1999: not sure how to properly check for this. Tried
```
      if(targetElementProps.classFullName !== GeometricElement3d.classFullName) {
        throw new IModelError(
          IModelStatus.WrongElement,
          `Expected GeometricElement3dProps, got ${targetElementProps.classFullName}`
        );
```
but threw
`Wrong Element: Expected GeometricElement3dProps, got Generic:PhysicalObject`
Is there a correct way to go about this. Is checking the instance of the target element (similar to how we check the src) enough?
`sourceElement instanceof GeometricElement3d`

tcobbs-bentley: I see that `GeometricElement3dProps` is an interface, so `instanceof` won't work.

DanRod1999: this: `targetElementProps instanceof GeometricElement3dProps`
gives the error
'GeometricElement3dProps' only refers to a type, but is being used as a value here.ts(2693)

tcobbs-bentley: You can create a function named `isGeometricElement3dProps` using Typescript [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates). The function will be more complicated than their example, but it shouldn't be _too_ bad.

ben-polinsky: We've kinda run into this before and decided to check `z` on placement...

tcobbs-bentley: Actually, what you really want is to check that `placement` is defined, and that it implements `Placement3dProps`. Creating a type predicate for `isPlacement3dProps` would be significantly easier than what I suggested above.

DanRod1999: we do have an isPlacement3dProps, but in order to call it my targetElementProps does need a valid placement, so I can use that, but it would kinda just switch around the current logic. I'll look into it and see if one is more reasonable than the other

DanRod1999: @mindaugasdirg do you still want some kind of change for this?

mindaugasdirg: I think we can leave it with current checks. My worry was that somebody will remap GeometricElement3d to non geometric element, but in real scenario that shouldn't be a valid remap because of other issues.

PR #265, opened 2025-07-09
Allow Performing Transforms with additionalTransforms

DanRod1999:
If src and target imodel's have differing local Helmert transforms, transform target elements positions to maintain proper local positions.

The target imodel "additionalTransform" will always be maintained, even if it is null. The source imodel when copied into the target will have a combination of both helmert transforms applied to its elements placements. This way the result will place the elements in a position that when the targets "additionalTransform" is applied it will be in the correct geographic position. If the target's "additionalTransform" is null, then the we are simply applying the src's helmert transform permanently to each elements position instead of just at render time. If the src and target have different non null transforms, then the resulting transform applied to the src elements will also negate the helmert transform applied by the target model.

### DIFF
--- a/packages/transformer/src/IModelTransformer.ts
+++ b/packages/transformer/src/IModelTransformer.ts
@@ -694,22 +694,26 @@ export class IModelTransformer extends IModelExportHandler {
     /* eslint-enable @itwin/no-internal */
     if (this._options.tryAlignGeolocation) {
       if (
-        this.sourceDb.geographicCoordinateSystem === undefined &&
-        this.targetDb.geographicCoordinateSystem === undefined
+        this.sourceDb.geographicCoordinateSystem ||
+        this.targetDb.geographicCoordinateSystem
       ) {
-        Logger.logTrace(
-          loggerCategory,
-          "Aligning ECEF Location's between imodels due to imodels not containing GeographicCoordinateSystem data"
-        );
-        this._linearSpatialTransform = this.calculateEcefTransform();
-      } else {
         Logger.logTrace(
           loggerCategory,
           "Aligning Additional transforms between imodels due to imodels containing GeographicCoordinateSystem data"
         );
         this._linearSpatialTransform =
           this.calculateSpatialTransfromFromHelmertTransforms();
-      }
+      } else if (this.sourceDb.ecefLocation && this.targetDb.ecefLocation) {
+        Logger.logTrace(
+          loggerCategory,
+          "Aligning ECEF Location's between imodels due to imodels not containing GeographicCoordinateSystem data"
+        );
+        this._linearSpatialTransform = this.calculateEcefTransform();
+      } else
+        Logger.logWarning(
+          loggerCategory,
+          "No Geolcation data to align, both GCS and ECEF are undefined"
+        );
     }
   }
 
--- a/packages/transformer/src/IModelTransformer.ts
+++ b/packages/transformer/src/IModelTransformer.ts
@@ -23,7 +23,14 @@ import {
   YieldManager,
 } from "@itwin/core-bentley";
 import * as ECSchemaMetaData from "@itwin/ecschema-metadata";
-import { Point3d, Transform } from "@itwin/core-geometry";
+import {
+  Angle,
+  AxisIndex,
+  Matrix3d,
+  Point3d,
+  Transform,
+  Vector3d,
+} from "@itwin/core-geometry";
 import * as coreBackendPkgJson from "@itwin/core-backend/package.json";
 import {
   BriefcaseManager,
@@ -82,6 +89,7 @@ import {
   FontProps,
   GeometricElement3dProps,
   GeometricElementProps,
+  Helmert2DWithZOffset,
   IModel,
   IModelError,
   ModelProps,
@@ -240,11 +248,18 @@ export interface IModelTransformOptions {
    * @default undefined
    */
   argsForProcessChanges?: ProcessChangesOptions;
-  /** A flag that determines if spatial elements from the source db should be transformed if source and target iModel ECEF locations differ.
+  /**
+   * A flag that determines if spatial elements from the source db should be transformed if source and target iModel ECEF locations differ.
    * @note This flag should only be used if imodels are linearly located
    * @default false
    */
   alignECEFLocations?: boolean;
+  /**
+   * A flag that determines if spatial elements from the source db should be transformed if source and target iModel GCS/CRS data is the same, but they have differing additional transforms.
+   * @note This flag should only be used if imodels are not linearly located
+   * @default false
+   */
+  alignAdditionalTransforms?: boolean;
 }
 
 /**
@@ -385,10 +400,14 @@ export class IModelTransformer extends IModelExportHandler {
   public readonly targetDb: IModelDb;
   /** The IModelTransformContext for this IModelTransformer. */
   public readonly context: IModelCloneContext;
-  /** The transform to be applied to the placement of spatial elements when source and target db have different ECEF locations
-   * @note transform can only be used when source and target are linearly located imodels
+  /** The transform to be applied to the placement of spatial elements
+   * This transform should be applied when:
+   * - source and target db have different ECEF locations
+   * - source and target db have matching GCS/CRS data, but differing `geographicCoordinateSystem.additionalTransform.helmert2DWithZOffset`
+   * @note for ECEF transforms, this can only be used when source and target are linearly located imodels
+   * @note for non linearly located imodels, this transform will be a linear transform derived from Helmert Transforms from the src and target iModels.
    */
-  public ecefTransform?: Transform;
+  public linearSpatialTransform?: Transform;
   private _syncType?: SyncType;
 
   /** The Id of the Element in the **target** iModel that represents the **source** repository as a whole and scopes its [ExternalSourceAspect]($backend) instances. */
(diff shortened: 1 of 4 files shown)


DanRod1999:
Still working on writing tests, but got this working in a POC. However, would like to mention how we could do this with a different approach.

Currently this process: 
takes a src imodel and then permanently moves its elements that are usually applied at render time. This removes the need for the `additionalTransfrom` to be included in the target imodel. Then another transform is applied that moves all the elements in the src such that when the targets `additionalTransform` is applied at render time the src elements are in the correct place. I combine these two transforms so its really only one mathematical transformation that is being applied, but the idea of moving an imodel just so at render time it can be move back seems silly.

That being said, the only way I can think of that avoids this is to: 
- preform place transform on src and apply its Helmert transform to all elements and remove the need for the `additionalTransform`
- Repeat for target
- Transform src into target now with no `additionalTransforms`.
This is a fair bit more work computationally i think, and certainly slower, since now we are shifting many more elements. However we are not doing any redundant transforms in the process, and no longer preforming any changes at render time. And technically little to no extra changes to the transformer would be needed.

I guess this is really only a problem if preforming the `additionalTransforms` are costly (which I assume they are not?). But if they are then we might want to go with this second workflow.

DanRod1999:
> @EarlinLutz and @dassaf4 for some math experience

Dave did helped me convert the helmert obj to a transform and helped verify it was producing the expected origin and coeffs values

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1651,6 +1688,97 @@ export class IModelTransformer extends IModelExportHandler {
     }
   }
 
+  public convertHelmertToTransform(
```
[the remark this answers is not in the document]
dassaf4: Sure thing. We've been working together on this, but I'll double check these methods now that we're out of draft.

dassaf4 (dismissed):

packages/transformer/src/IModelTransformer.ts
```diff
@@ -385,10 +400,14 @@ export class IModelTransformer extends IModelExportHandler {
...
-   */
-  public ecefTransform?: Transform;
+  /** The transform to be applied to the placement of spatial elements
+   * This transform should be applied when:
+   * - source and target db have different ECEF locations
+   * - source and target db have matching GCS/CRS data, but differing `geographicCoordinateSystem.additionalTransform.helmert2DWithZOffset`
+   * @note for ECEF transforms, this can only be used when source and target are linearly located imodels
+   * @note for non linearly located imodels, this transform will be a linear transform derived from Helmert Transforms from the src and target iModels.
+   */
+  public linearSpatialTransform?: Transform;
```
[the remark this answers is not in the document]
aruniverse: there used to be a `public` property `ecefTransform`, but that has now been renamed
that itself could be breaking
should also be called outin changelogs

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1651,6 +1691,97 @@ export class IModelTransformer extends IModelExportHandler {
...
+
+    if (
+      srcDb.geographicCoordinateSystem.additionalTransform?.helmert2DWithZOffset
+        ?.scale !==
+      targetDb.geographicCoordinateSystem.additionalTransform
+        ?.helmert2DWithZOffset?.scale
+    ) {
+      throw new IModelError(
+        IModelStatus.MismatchGcs,
+        "Spatial transform is non linear. Source and target Helmert transforms must have the same scale to calculate a linear spatial transform."
```
ben-polinsky, line 1760: Is this true? I thought scale was a linear transformation.

packages/transformer/src/test/standalone/GeoLocationTransform.test.ts
```diff
@@ -191,3 +213,139 @@ describe("Linear Geolocation Transformations", () => {
     transfrom.dispose();
   });
 });
+
+describe("Non Linear Geolocation Transformations", () => {
```
ben-polinsky, line 217: Might be good to refactor the test so you only have to provide two sets of transforms data to a helper function, and that way you can validate many transforms without replicating the whole test over and over. 

Definitely a bunch of bad paths to test as well.

packages/transformer/src/IModelTransformer.ts
dassaf4: When the spatial transform ends up with some scale, it's no longer a rigid transformation.  That's perfectly fine for a Transform, but I guess the iModel transformer doesn't allow non-rigid Transforms.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -1651,6 +1691,97 @@ export class IModelTransformer extends IModelExportHandler {
...
+    const helmertTransform = Transform.createRefs(translation, rotationXY);
+
+    return helmertTransform;
+  }
+
+  public calculateSpatialTransfromFromHelmertTransforms(
+    srcDb: IModelDb,
+    targetDb: IModel
+  ): Transform {
+    if (
```
mindaugasdirg, line 1720: The first two checks also force imodel-transformer consumers to do pretty much identical checks to determine which options can be passed to imodel-transformer if user wants to use the new feature.

```ts
// example usage of alignAdditionalTransforms and alignECEFLocations options
function runTransform(sourceDb: IModelDb, targetDb: IModelDb) {
    let alignAdditionalTransforms = false;
    let alignECEFLocations = false;

    if (sourceDb.geographicCoordinateSystem?.horizontalCRS !== undefined &&
        sourceDb.geographicCoordinateSystem?.verticalCRS !== undefined &&
        targetDb.geographicCoordinateSystem?.horizontalCRS !== undefined &&
        targetDb.geographicCoordinateSystem?.verticalCRS !== undefined) {
        // both use custom CRS
        alignAdditionalTransforms = true;
    } else if (sourceDb.ecefLocation !== undefined && targetDb.ecefLocation !== undefined) {
        alignECEFLocations = true;
    }
    )
    const transformer = new IModelTransformer(sourceDb, targetDb, {
        // ...other options...
        alignAdditionalTransforms,
        alignECEFLocations
    });

    await transformer.process();
}
```

I wonder if this is a good approach or a different approach would be better. For example, having a boolean option `tryAlignLocation` and IModelTransformer would execute the same checks and determine itself if this should be ECEF location adjustment, Helmert transform adjustment, log a message indicating which scenario was used andf or unsupported cases it would still throw an error to not mess up iModel data.

What do you think?

DanRod1999: I agree too. I think this makes more sense. Since you can only align ecef when both imodels are linearly located, it does not really make sense to give people an option. Plus it will prevent people from enabling both which doesn't make sense either. My only concern would be if in the future there are other situations where we want to align imodels in different ways, and in those cases there should be an option for the users to choose how they want to align them. However, I have no idea if this would ever actually happen, and I think if it does we could still work around it pretty easily

packages/transformer/src/IModelTransformer.ts
```diff
@@ -667,9 +692,31 @@ export class IModelTransformer extends IModelExportHandler {
...
+      ) {
+        Logger.logTrace(
+          loggerCategory,
+          "Aligning ECEF Location's between imodels due to imodels not containing GeographicCoordinateSystem data"
+        );
+        this._linearSpatialTransform = this.calculateEcefTransform(
+          this.sourceDb,
+          this.targetDb
+        );
+      } else {
```
mindaugasdirg, line 708: There is a third valid case when both iModels are not geolocated and both iModels have `geographicCoordinateSystem` and `ecefLocation` properties as undefined

DanRod1999: I do check for this in `calculateEcefTransform` and throw an error if ECEF location is missing. Is that enough, or would you rather have it caught here and the log a message instead of throwing an error?

packages/transformer/src/test/standalone/GeoLocationTransform.test.ts
ben-polinsky: We've discussed. Might not save any typing or time to extract, but go ahead and write some more cases and then maybe some helpers will emerge.

packages/transformer/src/IModelTransformer.ts
DanRod1999: I corrected this, thanks for the clarification

packages/transformer/src/IModelTransformer.ts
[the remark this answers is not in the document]
aruniverse: isnt this breaking? we previously had a public property, now you rename the property and make it private
shouldnt you deprecate the old property and point ppl to use the new one?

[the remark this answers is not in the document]
aruniverse: this change is making me want to enforce having api-extractor in this repo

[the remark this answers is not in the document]
aruniverse: ah if the entire class is beta, then this is fine

packages/transformer/src/IModelTransformer.ts
```diff
@@ -240,11 +248,21 @@ export interface IModelTransformOptions {
...
    */
   argsForProcessChanges?: ProcessChangesOptions;
-  /** A flag that determines if spatial elements from the source db should be transformed if source and target iModel ECEF locations differ.
+  /**
+   * A flag that determines if spatial elements from the source db should be transformed if source and target iModel ECEF locations differ.
    * @note This flag should only be used if imodels are linearly located
    * @default false
+   * @beta
+   */
+  // alignECEFLocations?: boolean;
```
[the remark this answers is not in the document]
DanRod1999: cleaned up comments

mindaugasdirg (approved):

ben-polinsky (approved):

PR #276, opened 2025-11-25
Allow to skip over reused element id as workaround for faulty changeset

khanaffan:


JulijaRamoskiene:
PR with these channges was created against master branch: https://github.com/iTwin/imodel-transformer/pull/279

PR #278, opened 2025-11-25
allow skip over reused ids

DanRod1999:
Cherry pick of [PR](https://github.com/iTwin/imodel-transformer/pull/276)
Resolves [issue](https://github.com/iTwin/itwinjs-backlog/issues/1762#issuecomment-3563326039)


DanRod1999:
Currently the test I've labeled with it.only is failing. Everything else succeeded. Have not yet figured out if it is due to bug in Affan's changes or if failure should be expected.

PR #279, opened 2025-11-25
Allow skip over reused ids

DanRod1999:
Cherry pick of https://github.com/iTwin/imodel-transformer/pull/276
Resolves https://github.com/iTwin/itwinjs-backlog/issues/1762#issuecomment-3563326039

### DIFF
--- a/packages/transformer/src/IModelExporter.ts
+++ b/packages/transformer/src/IModelExporter.ts
@@ -10,7 +10,6 @@ import {
   BriefcaseDb,
   BriefcaseManager,
   ChangedECInstance,
-  ChangesetECAdaptor,
   DefinitionModel,
   ECSqlStatement,
   // eslint-disable-next-line @typescript-eslint/no-redeclare
@@ -24,7 +23,6 @@ import {
   IModelHost,
   IModelJsNative,
   Model,
-  PartialECChangeUnifier,
   RecipeDefinitionElement,
   Relationship,
   SqliteChangeOp,
@@ -1170,7 +1168,33 @@ export class ChangedInstanceIds {
    * from the set of updatedIds and add it to the set of deletedIds for the appropriate class type.
    * @param change ChangedECInstance which has the ECInstanceId, changeType (insert, update, delete) and ECClassId of the changed entity
    */
-  public async addChange(change: ChangeInstanceKey): Promise<void> {
+  public async addChange(change: ChangedECInstance): Promise<void> {
+    if (!this._ecClassIdsInitialized) await this.setupECClassIds();
+    const ecClassId = change.ECClassId ?? change.$meta?.fallbackClassId;
+    if (ecClassId === undefined)
+      throw new Error(
+        `ECClassId was not found for id: ${change.ECInstanceId}! Table is : ${change?.$meta?.tables}`
+      );
+    const changeType: SqliteChangeOp | undefined = change.$meta?.op;
+    if (changeType === undefined)
+      throw new Error(
+        `ChangeType was undefined for id: ${change.ECInstanceId}.`
+      );
+    if (this._relationshipSubclassIdsToSkip?.has(ecClassId)) return;
+
+    if (this.isRelationship(ecClassId))
+      this.handleChange(this.relationship, changeType, change.ECInstanceId);
+    else if (this.isCodeSpec(ecClassId))
+      this.handleChange(this.codeSpec, changeType, change.ECInstanceId);
+    else if (this.isAspect(ecClassId))
+      this.handleChange(this.aspect, changeType, change.ECInstanceId);
+    else if (this.isModel(ecClassId))
+      this.handleChange(this.model, changeType, change.ECInstanceId);
+    else if (this.isElement(ecClassId))
+      this.handleChange(this.element, changeType, change.ECInstanceId);
+  }
+
+  public async addChangeKey(change: ChangeInstanceKey): Promise<void> {
     if (!this._ecClassIdsInitialized) await this.setupECClassIds();
     const ecClassId = change.classId;
     if (ecClassId === undefined)
@@ -1532,7 +1556,7 @@ class ChangesetProcessor {
           const key = makeKey(row);
           if (!instanceKeySet.has(key)) {
             instanceKeySet.add(key);
-            await store.addChange(row);
+            await store.addChangeKey(row);
           }
         }
       }
--- a/packages/transformer/src/IModelExporter.ts
+++ b/packages/transformer/src/IModelExporter.ts
@@ -31,6 +31,7 @@ import {
 import {
   assert,
   DbResult,
+  GuidString,
   Id64,
   Id64Arg,
   Id64Set,
@@ -1485,6 +1486,10 @@ interface ChangeInstanceKey {
   fallbackClassId: Id64String;
   /** The previous ECClassId before the change, if applicable */
   previousClassId?: Id64String;
+  /** The federation GUID associated with the instance, if available */
+  federationGuid?: GuidString;
+  /** The previous federation GUID associated with the instance, if available */
+  previousFederationGuid?: GuidString;
 }
 
 type ActionOnIdReuseDetected = "Skip" | "Fail";
@@ -1542,10 +1547,20 @@ class ChangesetProcessor {
       db: this.db,
       disableSchemaCheck: true,
     });
-
+    let fedGuidColumnIndex: number | undefined;
     try {
       while (csReader.step()) {
-        const row = this.processRow(csReader);
+        if (
+          csReader.tableName === "bis_Element" &&
+          fedGuidColumnIndex === undefined
+        ) {
+          fedGuidColumnIndex = csReader
+            .getColumnNames(csReader.tableName)
+            .indexOf("FederationGuid");
+          if (fedGuidColumnIndex < 0) fedGuidColumnIndex = undefined;
+        }
+
+        const row = this.processRow(csReader, fedGuidColumnIndex);
         // SKIP over row where ClassId was reused
         if (row) {
           if (row.isIdReused)
@@ -1579,7 +1594,8 @@ class ChangesetProcessor {
    * available in the changeset.
    */
   private processRow(
-    reader: SqliteChangesetReader
+    reader: SqliteChangesetReader,
+    fedGuidColumnIndex?: number
   ): ChangeInstanceKey | undefined {
     const tableName = reader.tableName;
     if (
@@ -1643,6 +1659,25 @@ class ChangesetProcessor {
       }
     }
 
+    let federationGuid: GuidString | undefined;
+    let previousFederationGuid: GuidString | undefined;
+    if (fedGuidColumnIndex) {
+      const oldFedGuid =
+        reader.op !== "Inserted"
+          ? reader.getChangeValueBinary(fedGuidColumnIndex, "Old")
+          : undefined;
+      const newFedGuid =
+        reader.op !== "Deleted"
+          ? reader.getChangeValueBinary(fedGuidColumnIndex, "New")
+          : undefined;
+      if (newFedGuid) {
+        federationGuid = ChangesetProcessor.convertBinaryToGuid(newFedGuid);
+      } else if (oldFedGuid) {
+        previousFederationGuid =
+          ChangesetProcessor.convertBinaryToGuid(oldFedGuid);
+      }
+    }
+
     return {
       instanceId,
       classId,
@@ -1651,7 +1686,21 @@ class ChangesetProcessor {
       op: reader.op,
       isIdReused,
       tableName,
+      federationGuid,
+      previousFederationGuid,
     };
+  }
+  private static convertBinaryToGuid(binaryGUID: Uint8Array): GuidString {
+    // Check if the array has 16 elements
+    if (binaryGUID.length !== 16) {
+      throw new Error("Invalid array length for Guid");
+    }
+    // Convert each element to a two-digit hexadecimal string
+    const hex = Array.from(binaryGUID, (byte) =>
+      byte.toString(16).padStart(2, "0")
+    );
+    // Join the hexadecimal strings and insert hyphens
+    return `${hex.slice(0, 4).join("")}-${hex.slice(4, 6).join("")}-${hex.slice(6, 8).join("")}-${hex.slice(8, 10).join("")}-${hex.slice(10, 16).join("")}`;
   }
   /**
    * Retrieves the ECClassId for a given instance directly from the database.
--- a/packages/transformer/src/IModelExporter.ts
+++ b/packages/transformer/src/IModelExporter.ts
@@ -1501,6 +1501,7 @@ type ActionOnIdReuseDetected = "Skip" | "Fail";
 class ChangesetProcessor {
   private _cacheTables = new Map<string, TableInfo>();
   private _onIdReuseDetected: ActionOnIdReuseDetected = "Fail";
+  private _deletedReusedIds: Set<GuidString> = new Set();
   public constructor(public readonly db: IModelDb) {}
 
   /**
@@ -1561,7 +1562,7 @@ class ChangesetProcessor {
         }
 
         const row = this.processRow(csReader, fedGuidColumnIndex);
-        // SKIP over row where ClassId was reused
+        // where ClassId is reused add to deletedReusedIds list and set op to inserted
         if (row) {
           if (row.isIdReused)
             if (this._onIdReuseDetected === "Fail") {
@@ -1672,7 +1673,7 @@ class ChangesetProcessor {
           : undefined;
       if (newFedGuid) {
         federationGuid = ChangesetProcessor.convertBinaryToGuid(newFedGuid);
-      } else if (oldFedGuid) {
+      } else if (oldFedGuid && oldFedGuid.length !== 0) {
         previousFederationGuid =
           ChangesetProcessor.convertBinaryToGuid(oldFedGuid);
       }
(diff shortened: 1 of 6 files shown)


DanRod1999:
Currently the test I've labeled with it.only is failing. Everything else succeeded. Have not yet figured out if it is due to bug in Affan's changes or if failure should be expected.

packages/transformer/src/IModelExporter.ts
```diff
@@ -1427,25 +1448,270 @@ export class ChangedInstanceIds {
...
+  /**
+   * Processes a single row from the changeset reader and extracts change information.
+   * @param reader The SqliteChangesetReader positioned at the current row.
+   * @returns A ChangeInstanceKey object containing the extracted change information, or undefined
+   * if the row should be skipped (e.g., system tables, invalid primary keys).
+   * @remarks This method handles special cases such as virtual ClassId columns, ClassId changes
+   * during updates (ID reuse), and falls back to querying the database when ClassId is not
+   * available in the changeset.
+   */
+  private processRow(
```
JulijaRamoskiene, line 1627: `processRow()` is re-implementing first half of `ChangesetECAdaptor.step()` with slight twists. 
Is it possible to add this fix into  `ChangesetECAdaptor` instead, it could be: 
- Optional step() option to skip parsing property changes
- Separate method that would be used to only extract changed element data without parsing properties
- or maybe there is a way to adjust ChangesetECAdaptorso so that this 'corrupted' change is parsed as  2 changes (delete and insert)? 

I think it is not best idea to put changeset parsing / reading logic  into this imodel-transformer repo, as:
- Code is duplicated
- Current implementation has a lot of internal details about parsing change sets. Transformer should not be aware of them. 
- It might  be hard to maintain and fix this code in case it fails, as transformer maintainers do not posses required knowledge about parsing change sets.
- If it is put here, it will fix issue only for our service. e.g. Changed elements service is also failing while processing such change set.
- We would need to write tests for change set processing here or copy some  unit tests from ChangesetECAdaptor.

packages/transformer/src/IModelExporter.ts
```diff
@@ -1427,25 +1448,270 @@ export class ChangedInstanceIds {
...
+  classId?: Id64String;
+  /** Indicates if the instance ID was reused (i.e., ClassId changed during an update operation) */
+  isIdReused?: true;
+  /** The type of change operation performed on the instance */
+  op: SqliteChangeOp;
+  /** The name of the ECDb table containing the changed instance */
+  tableName: string;
+}
+
+type ActionOnIdReuseDetected = "Skip" | "Fail";
```
JulijaRamoskiene, line 1509: Both actions in  `ActionOnIdReuseDetected` fail to  handle both elements. 

 "Skip" option means that in scenario where element is deleted, and other element with same instance id is added transformer will:
- leave old deleted element in target
- add new element to source

Old deleted element needs to be deleted from target.

packages/transformer/src/IModelExporter.ts
```diff
@@ -1427,25 +1448,270 @@ export class ChangedInstanceIds {
...
+      await this.processFile(csFileProp, store);
+    }
+  }
+
+  /**
+   * Processes a single changeset file and extracts changed instance information.
+   * @param csFileProp The changeset file properties.
+   * @param store The ChangedInstanceIds object to populate with extracted changes.
+   * @remarks This method reads through the changeset file, processes each row, and adds valid changes
+   * to the store. Duplicate changes (based on instanceId and classId) are skipped, and rows where
```
JulijaRamoskiene, line 1549: It might not be safe to skip changes: What if element is deleted then re-added (with same class)? 
In case this skipping logic is needed, consider adding operation type into key in `makeKey()`

packages/transformer/src/IModelExporter.ts
JulijaRamoskiene: Note: transformer currently does not support adding same instance id as deleted &  inserted / updated: https://github.com/iTwin/imodel-transformer/blob/df016da7573c13c5d45e7aabbdf3e2c7be045e13/packages/transformer/src/IModelExporter.ts#L1361

JulijaRamoskiene (changes requested):

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -1412,7 +1412,7 @@ describe("IModelTransformerHub", () => {
     await tearDown();
   });
 
-  it("should merge changes made on a branch back to master", async () => {
+  it.only("should merge changes made on a branch back to master", async () => {
```
JulijaRamoskiene: Processinng user iModel fails with same error as `should merge changes made on a branch back to master` test.
Unit test failure: 
```
1) IModelTransformerHub
        should merge changes made on a branch back to master:
      Error: ECClassId was not found for id: 0x17! Table is : bis_Element
       at ChangedInstanceIds.addChangeKey (src\IModelExporter.ts:1201:13)
       at ChangesetProcessor.processFile (src\IModelExporter.ts:1559:25)
       at ChangesetProcessor.processFiles (src\IModelExporter.ts:1517:7)
       at Function.initialize (src\IModelExporter.ts:1455:5)
       at IModelExporter.initialize (src\IModelExporter.ts:363:29)
       at IModelTransformer.initialize (src\IModelTransformer.ts:3009:5)
       at IModelTransformer.process (src\IModelTransformer.ts:3406:5)
       at runTimeline (src\test\TestUtils\TimelineTestUtil.ts:527:11)
       at Context.<anonymous> (src\test\standalone\IModelTransformerHub.test.ts:1676:42) 
 
```

Exception thrown while processing user iModel:

```
ECClassId was not found for id: 0x470000000e2a! Table is : bis_Element | stack: Error: ECClassId was not found for id: 0x470000000e2a! Table is : bis_Element\n
    at ChangedInstanceIds.addChangeKey (C:\\repositories\\git\\iModelTransformations\\agent\\dist\\main.js:328374:19)\n
    at ChangesetProcessor.processFile (C:\\repositories\\git\\iModelTransformations\\agent\\dist\\main.js:328643:37)\n    
	at async ChangesetProcessor.processFiles (C:\\repositories\\git\\iModelTransformations\\agent\\dist\\main.js:328609:13)\n
    at async ChangedInstanceIds.initialize (C:\\repositories\\git\\iModelTransformations\\agent\\dist\\main.js:328577:9)\n
    at async Exporter.initialize (C:\\repositories\\git\\iModelTransformations\\agent\\dist\\main.js:327716:33)\n    
	at async Transformer.initialize (C:\\repositories\\git\\iModelTransformations\\agent\\dist\\main.js:109831:9)\n    
	at async Transformer.initialize (C:\\repositories\\git\\iModelTransformations\\agent\\dist\\main.js:168622:9)\n    
	at async descriptor.value (C:\\repositories\\git\\iModelTransformations\\agent\\dist\\main.js:269630:32)\n    
	at async Transformer.manualProcessChangesInitialize (C:\\repositories\\git\\iModelTransformations\\agent\\dist\\main.js:169122:9)\n    
	at async Transformer.startTransformation (C:\\repositories\\git\\iModelTransformations\\agent\\dist\\main.js:168512:13)
```

packages/transformer/src/IModelExporter.ts
khanaffan: This isn’t really a fix—it’s more of a hack/workaround. Changing our API to handle incorrect data could cause bigger problems later. It’s not just the transformer; the change element API will also be impacted, and version comparisons might fail. This introduces a new scenario that everyone would need to account for. The real issue is that the data is wrong, not that our APIs are lacking.

I suggest we fix the connector and, if possible, delete the changeset and regenerate the data. Otherwise, this incorrect data could cause long-term headaches.

khanaffan: I will add check on imodel02 and itwin.js that fail to push changeset where ECClassId was updated. This will fail the connector or any client that attempt to push data with updated ECClassId.

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
khanaffan: Fixed

```cs
│   imodel-transformer
│     ✔ should should simplify Element geometry in the target iModel (387ms)
│     ✔ should should combine PhysicalModels in the target iModel (281ms)
│     ✔ should exclude categories (315ms)
│     ✔ should clone element struct array entries including when their class layout changes (516ms)
│     ✔ should clone element structs values (476ms)
│   5 passing (2s)
```

packages/transformer/src/IModelExporter.ts
khanaffan: For given changeset a row in table can only appear once as either inserted or update or deleted. OP will have no effect. Though i my recent fix for test i expanded the key to include classid, fallbackclassid as classid can be undefined.

packages/transformer/src/IModelExporter.ts
khanaffan: That is correct. @DanRod1999 can we do that?. Can we create list of element that we want to delete before starting processing changed elementids.

khanaffan: I also agree with @JulijaRamoskiene  that simply skipping is not correct behaviour. DELETE and insert is the correct solution. For that I have provided enough information from ChangesetProcessor that should help @DanRod1999 make a change down in transformer to delete and insert. There should be seperate list of deleted elements that had reused the id and they should be recorded as insert in the changeinstanceids data structure.

DanRod1999:
> I suggest we fix the connector and, if possible, delete the changeset and regenerate the data. Otherwise, this incorrect data could cause long-term headaches.

If we are able to do this, is this PR even worth attempting? It seems like we'd be including some fairly redundant code just for the sake of handling improper data?

packages/transformer/src/IModelExporter.ts
```diff
@@ -1427,25 +1450,337 @@ export class ChangedInstanceIds {
...
+        reader.op !== "Inserted"
+          ? reader.getChangeValueBinary(fedGuidColumnIndex, "Old")
+          : undefined;
+      const newFedGuid =
+        reader.op !== "Deleted"
+          ? reader.getChangeValueBinary(fedGuidColumnIndex, "New")
+          : undefined;
+      if (newFedGuid && newFedGuid.length !== 0) {
+        federationGuid = ChangesetProcessor.convertBinaryToGuid(newFedGuid);
+      } else if (oldFedGuid && oldFedGuid.length !== 0) {
```
JulijaRamoskiene: [Q] Why `else if`? Why `previousFederationGuid` is parsed only if `newFedGuid` is empty?

JulijaRamoskiene: @DanRod1999 since `previousFederationGuid` is almost never parsed,  code below  fails to collect deleted elements:
```typescript
        // where ClassId is reused add to deletedReusedIds list
        if (row) {
          if (row.isIdReused && row.previousFederationGuid) {
            store.deletedReusedIds.add(row.previousFederationGuid);
            row.op = "Inserted";
          }
```

DanRod1999: @khanaffan, is my logic for creating a list of deleted elements wrong here, or as Julija mentioned, should previous fedGuid be parsed regardless of newFedGuid's contents?

khanaffan: > [Q] Why `else if`? Why `previousFederationGuid` is parsed only if `newFedGuid` is empty?

fixed.

khanaffan: > be parsed regardless of newFedGuid's contents?

seem fine. I am not sure if row.op is used but if its used downstream then it should be row.op = "Deleted".

DanRod1999:
Was able to test some of these changes in transformation service, and intial results look promising, however further testing is needed. I was able to get past the error the service transformer is hitting, and it looks like duplicate elements are not appearing. Currently attempting to adjust Affan's test "Instance update to a different class (bug)" to process changesets with new transformer logic instead of ChangesetECAdapter. Hopefully I can see the same behavior, and then have an easier test bed to check if Transformer logic to delete duplicate entities is working correctly.

DanRod1999:
I updated the test to perform a transformation that processes changes of the buggy changeset created in the test. Behavior seems to be correct but if someone can review my process and checks appear correct that would be great. I'm currently trying to figure out how to expand this test to also check models and relationships with improper reused ids.

I'd like to clarify one thing though. I am currently using imodel importer's delete which uses deleteElmeentTreeCascade(). My understanding is this is so we delete all related elements part of its tree. However, is it possible that this would delete elements that are still relevant to the element that still exists in the imodel. For example, lets say I have el1 that goes through a bad changeset and that instanceId is replaced with el2. If el1 has a child el1child, how do we know that we want to delete that child element and that its no longer supposed to be related to el2? In the case that el1child had a similar issue and was replaced with el2child than that element should also hit the reusedId logic and be deleted. Is it guaranteed that we want to delete all related elements?

packages/transformer/src/test/standalone/BadChangeSet.test.ts
```diff
@@ -0,0 +1,421 @@
...
+  });
+
+  it("Instance update to a different class (bug)", async () => {
+    /**
+     * Test scenario: Verifies changeset reader behavior when an instance ID is reused with a different class.
+     *
+     * Steps:
+     * 1. Import schema with two classes (T1 and T2) that inherit from GraphicalElement2d.
+     *    - T1 has property 'p' of type string
+     *    - T2 has property 'p' of type long
```
JulijaRamoskiene, line 51: Have you tried setting property types to something that would throw `InvalidRange` error? Could be a separate test to make sure our parsers are not throwing.

DanRod1999: I have not, but I'll try

mindaugasdirg:
> I updated the test to perform a transformation that processes changes of the buggy changeset created in the test. Behavior seems to be correct but if someone can review my process and checks appear correct that would be great. I'm currently trying to figure out how to expand this test to also check models and relationships with improper reused ids.
> 
> I'd like to clarify one thing though. I am currently using imodel importer's delete which uses deleteElmeentTreeCascade(). My understanding is this is so we delete all related elements part of its tree. However, is it possible that this would delete elements that are still relevant to the element that still exists in the imodel. For example, lets say I have el1 that goes through a bad changeset and that instanceId is replaced with el2. If el1 has a child el1child, how do we know that we want to delete that child element and that its no longer supposed to be related to el2? In the case that el1child had a similar issue and was replaced with el2child than that element should also hit the reusedId logic and be deleted. Is it guaranteed that we want to delete all related elements?

No, deleteElementTreeCascade deletes assumes that all related elements should also be deleted. Previously, it worked because transformer had already iterated through all the source iModel entities and all the parent-child, model-element relationships would have been updated and wouldn't have relationship to the deleted element. For example, during the first transformation source iModel had elements el1 and el2 and el1 element had child elements el11 and el12. 

```
Root subject -> el1 -> el11
         |       |
         |       ----> el12
         ---------> el2
```
Then the second transformation processes changesets where element el1 and el11 were deleted, while element el12 has been updated to have el2 as a parent.
```
Root subject -> el2 -> el12
```

Transformer will first iterate element el12, update it and only then it will process the deletes deleting el1 and el11 using deleteElementTreeCascade.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -3010,8 +3023,95 @@ export class IModelTransformer extends IModelExportHandler {
...
+                loggerCategory,
+                `Failed to delete model ${instanceToDelete.instanceId} from class ${classFullName}: ${error}`
+              );
+            }
+            break;
+        }
+      }
+    } else {
+      // Exporter must be initialized prior to processing changesets in order to properly handle entity recreations (an entity delete followed by an insert of that same entity).
+      await this.processChangesets();
```
DanRod1999, line 3113: Calling this function was throwing errors when the elements with reusedIds had class changes with conflicting types. You can see this test: "Instance update to a different class and prop types (bug)", to check how I did this, and confirm it is valid.

To solve this problem I wrapped the processChangesets() in and else stmt, so it should only run if no reused ids were caught. However, I am not confident that this is the correct solution.

packages/transformer/src/test/standalone/BadChangeSet.test.ts
DanRod1999: I added a second test to try this

packages/transformer/src/test/standalone/BadChangeSet.test.ts
```diff
@@ -0,0 +1,765 @@
...
+    });
+    // 1. Import schema with classes that span overflow table.
+    // Schema replicates the incompatibility seen between RevitDynamic:Gate_Curtain_Wall_Single (EDITED_BY as Timestamp/dateTime)
+    // and RevitDynamic:System_Panel (EDITED_BY as string)
+    const schema = `<?xml version="1.0" encoding="UTF-8"?>
+    <ECSchema schemaName="TestDomain" alias="ts" version="01.00" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.1">
+        <ECSchemaReference name="BisCore" version="01.00" alias="bis"/>
+        <ECEntityClass typeName="T1">
+            <BaseClass>bis:GraphicalElement2d</BaseClass>
+              <ECProperty propertyName="EDITED_BY" typeName="dateTime"/>
```
DanRod1999, line 478: This test mimics the conditions of the user error that caught this by using a class that has a date time type and a 2nd class that only has a string type.

This seems to reproduce the error we saw in user model, and when testing the user model with my changes to the transformer I did not get any invalid time errors. However, in this test I still get invalid time errors when preforming the transform. Because of this I changed some of the logic in imodelTransformer (see comment above). I need to retry my prev changes with the user model and confirm I do not get the invalid time error. If I do get a successful run I assume this test differs in some way, but I'm unsure how that would be.

PR #288, opened 2026-03-30
wip add option to skip exportAll()

### LINKED ISSUE #259: Should `exportChanges()` fall back to `exportChanges()` when source has no changesets?
`IModelExporter:exportChanges()` will fall back to exportAll() when source has no changesets. Suggest to reconsider if this is valid behavior, as it might be unexpected to some users. Maybe error or silent return would be best here. 
At least in our tests I found a bunch of cases that were meant to test change processing, but were really falling back to process all / export all workflow.

https://github.com/iTwin/imodel-transformer/blob/2508c4cfc49d73ba9a710d2e04a8feecf3e68cd6/packages/transformer/src/IModelExporter.ts#L441C1-L444C6

DanRod1999:
proposing solution for https://github.com/iTwin/imodel-transformer/issues/259

Currently default is set to true, if this behavior is acceptable we may want to make the default to false, which would be a breaking change behavior wise. Otherwise, if we don't want this option at all we should just remove.

### DIFF
--- a/packages/transformer/src/IModelExporter.ts
+++ b/packages/transformer/src/IModelExporter.ts
@@ -81,6 +81,7 @@ export type ExporterInitOptions = ExportChangesOptions;
  */
 export type ExportChangesOptions = {
   skipPropagateChangesToRootElements?: boolean;
+  exportAllIfNoChangesetsPresent?: boolean;
 } /**
  * an array of ChangesetFileProps which are used to read the changesets and populate the ChangedInstanceIds using [[ChangedInstanceIds.initialize]] in [[IModelExporter.exportChanges]]
  * @note mutually exclusive with @see changesetRanges, @see startChangeset and @see changedInstanceIds, so define one of the four, never more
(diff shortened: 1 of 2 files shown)


DanRod1999:
@iTwin/transformation-service, any opinions on what behavior we want to move forward with? If there is no reason for a fallback then I would like to remove this all together, but not sure if the fallback is needed for any specific use cases

JulijaRamoskiene:
I am also not sure for what purpose `exportChanges()` is falling back to `exportAll()`. 
It is not straightforward to tell when changes were processed and when all data was imported without closely investigating the target (checking elements that should be deleted). 
Users might think that they are doing change processing, when in reality due to bad configuration they are falling back to `exportAll`. I saw this in tests, where assumption was that change processing was being tested, and in reality they were false-positive due to `exportAll` fallback. 

I agree that this would be a breaking change, so it should be done with major version change and documentation. We could also review `processAll` vs `exportAll`. Maybe some code can be re-used there. 
 
I don't think it is a good idea to introduce option as it brings in more confusion and is not solving anything. If we are not ready to remove the fallback, then I think we should not introduce the option either.

Issue #236, opened 2025-02-19
First process changes transformation should not reprocess changesets that were already handled with process all.


JulijaRamoskiene:
[Synchronization version is not saved](https://github.com/iTwin/imodel-transformer/blob/4e4a7df0030ea7fba666c4b700dc6f4160074340/packages/transformer/src/IModelTransformer.ts#L2164) when initial process all transformation is executed. 

Meaning, that when first process changes transformation is started, it will not have information about previously processed changesets, and it will reprocess all changesets again. After all is done, it will save version at the end of transformation.

Second process changes transformation will find synchronization version and will process only new changesets as expected. 

# To Reproduce
 
```typescript
describe("Process changes", () => {
    let sourceDb: BriefcaseDb;
    let targetDb: BriefcaseDb;

    beforeEach(async () => {
      sourceDb = await prepareBriefcase("source");
      targetDb = await prepareBriefcase("target");
    });

    afterEach(async () => {
      await closeAndDeleteBriefcase(sourceDb);
      await closeAndDeleteBriefcase(targetDb);
    });

    it("should handle deletes when process changes", async () => {
      // set up source
      const sourceModelId0 = PhysicalModel.insert(sourceDb, IModel.rootSubjectId, "M0");
      PhysicalModel.insert(sourceDb, IModel.rootSubjectId, "M1");
      const categoryId: Id64String = SpatialCategory.insert(sourceDb, IModel.dictionaryId, "SpatialCategory", {})
      const codePhysical1 = new Code({ scope: "0x1", spec: "0x1", value: "PhysicalOne" });
      const physicalElement1: PhysicalElementProps = {
        classFullName: PhysicalObject.classFullName,
        model: sourceModelId0,
        category: categoryId,
        code: codePhysical1
      };
      const physicalElement1Id = sourceDb.elements.insertElement(physicalElement1);

      const codePhysical2 = new Code({ scope: "0x1", spec: "0x1", value: "PhysicalTwo" });
      const physicalElement2: PhysicalElementProps = {
        classFullName: PhysicalObject.classFullName,
        model: sourceModelId0,
        category: categoryId,
        code: codePhysical2
      };
      const physicalElement2Id = sourceDb.elements.insertElement(physicalElement2);

      await pushChanges(sourceDb, "Initial source data");
      
      // ========= Transformation 1: Initial process all transformation =================
      let transformer = new IModelTransformer(sourceDb, targetDb);
      await transformer.process();
[...]

PR #295, opened 2026-04-24
Breakup IModelTransformer class

### LINKED ISSUE #294: Break up IModelTransformer Class
IModelTransformer class is very large, and does not necessarily need to contain all that it does. 
Suggest classes to extract:
`SyncTypeResolver`
`ProvenanceManager`
`ChangeDetector`
`ElementResolver`

DanRod1999:
This was a suggestion by copilot and I think aligned a little with one of the comments in the service's teams issue about reducing size of some of the files. I reviewed this work as copilot was making the changes, and I feel its fine. However, I would like to get the perspective of the service team. These are pretty dramatic changes, and I don't necessarily know if I agree that its worthwhile.

Decompose IModelTransformer into focused internal classes

Extracts three internal classes from IModelTransformer (~3,500 → ~2,500 lines), each owning a single responsibility:

 | Class              | Responsibility                                                              | Lines  |
 |--------------------|-----------------------------------------------------------------------------|--------|
 | SyncTypeResolver   | Determines sync direction (forward/reverse/not-sync) by inspecting ESAs     | ~125   |
 | ProvenanceManager  | Manages scope ESA lifecycle, sync versioning, provenance queries, iteration | ~1,000 |
 | ElementResolver    | Resolves target element via remap → FedGuid → Code fallback                | ~115   |

Dependency Design

All three classes receive IModelCloneContext via composition. The dependency chain is:

 IModelCloneContext → SyncTypeResolver → ProvenanceManager
 IModelCloneContext → ElementResolver

ProvenanceManager computes _startingChangesetIndices from the context internally. IModelTransformer retains thin delegate methods for externally-used
APIs.

Breaking Changes

The following public/static members were removed from IModelTransformer:

 - determineSyncType() — moved to SyncTypeResolver.determineSyncType()
 - noEsaSyncDirectionErrorMessage — moved to SyncTypeResolver.noEsaSyncDirectionErrorMessage
 - queryScopeExternalSourceAspect() — moved to ProvenanceManager.queryScopeExternalSourceAspect()
 - getProvenanceDb() / getProvenanceSourceDb() — use _provenanceManager.getProvenanceDb() (now protected)
 - getIsReverseSynchronization() / getIsForwardSynchronization() — use _provenanceManager or _syncTypeResolver
 - forEachTrackedElement() (static) — moved to ProvenanceManager.forEachTrackedElement()
 - initElementProvenanceOptions() (static) — moved to ProvenanceManager.initElementProvenanceOptions()
 - initRelationshipProvenanceOptions() (static) — moved to ProvenanceManager.initRelationshipProvenanceOptions()

Retained delegates (used by imodel-transformations-service): initElementProvenance, getSynchronizationVersion, tryGetProvenanceScopeAspect, 
initScopeProvenance, updateSynchronizationVersion — one-liner pass-throughs on IModelTransformer.

_provenanceManager is now protected so subclasses can access the full ProvenanceManager API directly.

Bug Fix: allowNoScopingESA removal

Removed the private allowNoScopingESA hack from SyncTypeResolver. Two tests were using argsForProcessChanges for initial imports (invalid workflow —
should use processAll), then setting allowNoScopingESA = true to suppress the resulting error. The tests now correctly use processAll for the initial
import.

Related fix: processAll now explicitly sets _sourceChangeDataState = "has-changes" so that updateSynchronizationVersion writes the sync version after a
full export. A noProvenance guard was added to ProvenanceManager.updateSynchronizationVersion to prevent crashes when no scope ESA exists.

Test Improvements

 - Replaced brittle transformer["_provenanceManager"]["_targetScopeProvenanceProps"] access with ProvenanceManager.queryScopeExternalSourceAspect() 
database queries
 - Initial import tests now use the correct processAll workflow


khanaffan (approved):

packages/transformer/src/SyncTypeResolver.ts
```diff
@@ -0,0 +1,149 @@
...
+/** Options for constructing a SyncTypeResolver.
+ * @internal
+ */
+export interface SyncTypeResolverOptions {
+  sourceDb: IModelDb;
+  targetDb: IModelDb;
+  targetScopeElementId: Id64String;
+  isProvenanceInitTransform?: boolean;
+  allowNoScopingESA?: boolean;
+  hasArgsForProcessChanges?: boolean;
```
JulijaRamoskiene: - Who `hasArgsForProcessChanges` ? Name makes no sense in `SyncTypeResolver` class.
- Why all these values are put into options? Why those are not passed separately into constructor?
- Why  `determineSyncType()` is static and public? 

It seems that there are some rules how to use this class, but it is not intuitive to grasp from code / implementation. I see that `getResolvedSyncType()` throws if `resolve()` was not called before, but when is the right time to call resolve? Maybe better implementation would help to lift that requirement.

packages/transformer/src/ProvenanceManager.ts
```diff
@@ -0,0 +1,909 @@
...
+import type { Entity } from "@itwin/core-backend";
+import { TransformerLoggerCategory } from "./TransformerLoggerCategory";
+import type { TargetScopeProvenanceJsonProps } from "./IModelTransformer";
+
+const loggerCategory: string = TransformerLoggerCategory.IModelTransformer;
+
+/** Options for constructing a ProvenanceManager.
+ * @internal
+ */
+export interface ProvenanceManagerOptions {
```
JulijaRamoskiene: - Why all these properties are put under `ProvenanceManagerOptions` instead of passing them separately into ctor? 
- Why `ProvenanceManagerOptions` has `transformerOptions`, does it help to decouple implementation?
- Design and usage of  `getIsReverseSynchronization()` / `queryTargetRelationshipId()` feels over-complicated:

Image

packages/transformer/src/IModelTransformer.ts
```diff
@@ -680,6 +526,32 @@ export class IModelTransformer extends IModelExportHandler {
       (this.targetDb as any).codeValueBehavior = "exact";
     }
     /* eslint-enable @itwin/no-internal */
+    this._syncTypeResolver = new SyncTypeResolver({
```
JulijaRamoskiene: What is the plan for unit tests? Will it be easy to test / mock when values are not passed trough ctor?

JulijaRamoskiene (commented):
- Feel like classes are still tightly coupled. Left comments.
- No tests for new classes. Refactoring of old tests might be required.
- All classes have options, not sure why those are needed.

packages/transformer/src/SyncTypeResolver.ts
DanRod1999: I cleaned up this a bit. Removed options as I agree, they weren't necessary (I actually thought I removed them at first but what copilot actually did was remove move the options interface it created from imodelTransformer to the new files)

determineSyncType was always static and public I believe, so I kept it the same. Do you think it should change to private now that I moved it into its own class? I believe this design decision was originally made so users could potentially do this:
```typescript
  /** Create an ExternalSourceAspectProps in a standard way for an Element in an iModel --> iModel transformation. */
  public static initElementProvenanceOptions(
    sourceElementId: Id64String,
    targetElementId: Id64String,
    args: {
      sourceDb: IModelDb;
      targetDb: IModelDb;
      // TODO: Consider making it optional and determining it through ESAs if not provided. This gives opportunity for people to determine it themselves using public static determineSyncType function.
      isReverseSynchronization: boolean;
      targetScopeElementId: Id64String;
    }
  ): ExternalSourceAspectProps {
    return ProvenanceManager.initElementProvenanceOptions(
      sourceElementId,
      targetElementId,
      args
    );
  }
```

packages/transformer/src/IModelTransformer.ts
DanRod1999: my new impl should make make testing/mocking easy, but lmk if that's not what you were referring to

packages/transformer/src/ProvenanceManager.ts
DanRod1999: > Design and usage of getIsReverseSynchronization() / queryTargetRelationshipId() feels over-complicated

I didn't like this either, but these 2 functions depend on imodelTransformer context that the provenance manager doesn't have. I feel like queryTargerRelId could be moved to manager, but that function depends on the transformer context, which I don't think we want to pass to the manager.

Similar with getIsReverseSync, we would need to pass in the sync type resolver to the provenance manager

packages/transformer/src/IModelTransformer.ts
```diff
@@ -421,148 +408,12 @@ export class IModelTransformer extends IModelExportHandler {
     "targetScopeElementId" | "danglingReferencesBehavior"
   >;
 
-  /**
-   * A private variable meant to be set by tests which have an outdated way of setting up transforms. In all synchronizations today we expect to find an ESA in the branch db which describes the master -> branch relationship.
-   * The exception to this is the first transform aka the provenance initializing transform which requires that the master imodel and the branch imodel are identical at the time of provenance initialization.
-   * A couple ofoutdated tests run their first transform providing a source and targetdb that are slightly different which is no longer supported. In order to not remove these tests which are still providing value
-   * this private property on the IModelTransformer exists.
-   */
-  private _allowNoScopingESA = false;
```
JulijaRamoskiene, line 430: suggest to add _allowNoScopingESA documentation in SyncTypeResolver as it is valuable (was not aware of the purpose for this property until I read this). It would be good to check if any other valuable comments were trimmed in this PR.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -680,6 +526,30 @@ export class IModelTransformer extends IModelExportHandler {
...
+      this._options.targetScopeElementId,
+      this._options,
+      async () => this.getIsReverseSynchronization(),
+      this._startingChangesetIndices,
+      async (sourceRelInfo) => this._queryTargetRelId(sourceRelInfo)
+    );
+    this._elementResolver = new ElementResolver({
+      targetDb: this.targetDb,
+      isBetweenIModels: this.context.isBetweenIModels,
+      findTargetElementId: (sourceId) =>
```
JulijaRamoskiene: Suggest to pass IModelCloneContext  object instead of  2 callbacks in ctor : `findTargetElementId` & `remapElement`.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -680,6 +526,30 @@ export class IModelTransformer extends IModelExportHandler {
...
+      !!this._options.argsForProcessChanges
+    );
+    this._provenanceManager = new ProvenanceManager(
+      this.sourceDb,
+      this.targetDb,
+      this._options.targetScopeElementId,
+      this._options,
+      async () => this.getIsReverseSynchronization(),
+      this._startingChangesetIndices,
+      async (sourceRelInfo) => this._queryTargetRelId(sourceRelInfo)
```
JulijaRamoskiene: Suggest to move `_queryTargetRelId` to `ProvenanceManager`. IModelCloneContext would be passed to ctor instead of calllback

packages/transformer/src/IModelTransformer.ts
JulijaRamoskiene: `isBetweenIModels` would be redundant in that case as well.

packages/transformer/src/IModelTransformer.ts
```diff
@@ -2335,155 +1634,6 @@ export class IModelTransformer extends IModelExportHandler {
...
-   * @note If [[IModelTransformOptions.argsForProcessChanges]] is not defined in this transformation, this function will return early without updating the sync version,
-   * unless the `initializeReverseSyncVersion` option is set to `true`
-   *
-   * The `initializeReverseSyncVersion` is added to set the reverse synchronization version during a forward synchronization.
-   * When set to `true`, it saves the reverse sync version as the current changeset of the targetDb. This is typically used for the first transformation between a master and branch iModel.
-   * Setting `initializeReverseSyncVersion` to `true` has the effect of making it so any changesets in the branch iModel at the time of the first transformation will be ignored during any future reverse synchronizations from the branch to the master iModel.
-   *
-   * Note that typically, the reverseSyncVersion is saved as the last changeset merged from the branch into master.
-   * Setting initializeReverseSyncVersion to true during a forward transformation could overwrite this correct reverseSyncVersion and should only be done during the first transformation between a master and branch iModel.
```
JulijaRamoskiene, line 2354: one more comment that was removed

packages/transformer/src/SyncTypeResolver.ts
```diff
@@ -0,0 +1,139 @@
...
+          } else {
+            throw err;
+          }
+        }
+      }
+    }
+    return this._syncType;
+  }
+
+  public get allowNoScopingESA(): boolean {
```
JulijaRamoskiene: I don't think that these should be public, currently those were exposed only for tests. 
If we want to keep them, `allowNoScopingESA` setter should clear cached `_syncType ` on value change. 


No unit tests were added for any of the new classes.

JulijaRamoskiene: Alternative approach could be to mark `_syncTypeResolver`, `_provenanceManager` and `_elementResolver` as protected in `IModelTransformer`. And then create a special class only for tests where those could be passed into ctor: 

```
export class IModelTransformerForTests extends IModelTransformer {
    public constructor(
    source: IModelDb | IModelExporter,
    target: IModelDb | IModelImporter,
    options?: IModelTransformOptions,
    elementResolver?: ElementResolver,
    provenanceManager?: ProvenanceManager,
    syncTypeResolver?: SyncTypeResolver

    ) {
      super(source, target, options);
      this._elementResolver = elementResolver ?? this._elementResolver;
      this._provenanceManager = provenanceManager ?? this._provenanceManager;
      this._syncTypeResolver = syncTypeResolver ?? this._syncTypeResolver;
    }
```

Then in tests you could create `syncTypeResolver` with desired `allowNoScopingESA` value passed into ctor. 


This would "hide" ElementResolver, ProvenanceManager and SyncTypeResolver as internal details from users, but would still provide possibility to adjust these objects in tests.

JulijaRamoskiene: Suggestion inspired by https://github.com/iTwin/itwinjs-core/blob/master/core/ecschema-editing/src/Editing/Mutable/MutableSchema.ts

packages/transformer/src/IModelTransformer.ts
JulijaRamoskiene: We had internal discussion in our team, and generated suggestion. Added comment here: https://github.com/iTwin/imodel-transformer/pull/295#discussion_r3217841676

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
```diff
@@ -248,7 +248,9 @@ describe("IModelTransformerHub", () => {
         targetBriefcase
       );
```
JulijaRamoskiene, line 250: any plans for splitting / refactoring tests as well?

JulijaRamoskiene (changes requested):

packages/transformer/src/test/standalone/IModelTransformerHub.test.ts
DanRod1999: I looked into splitting tests, but not very closely. From what I've seen none of the current tests in imodelTransformer.test.ts or the hub tests focus specifically on syncTypeResolver and Provenance management. Instead they are more like integration tests. I do need to take a closer look, but that was my initial findings. Maybe it would be better to create new unit tests

packages/transformer/src/SyncTypeResolver.ts
DanRod1999: After looking into the `allowNoScopingESA` field more, I feel like it can be removed all together as I believe it doesn't reflect workflows that can exist. I have a proposed change that I believe works, but would need that to be reviewed by others to double check I am not breaking existing valid workflows. If my changes do seem wrong. I will proceed with this suggestion

packages/transformer/src/IModelTransformer.ts
DanRod1999: Trying to remove this all together so I will not add back this comment

packages/transformer/src/IModelTransformer.ts
```diff
@@ -3338,6 +2294,8 @@ export class IModelTransformer extends IModelExportHandler {
    * @note [[processSchemas]] is not called automatically since the target iModel may want a different collection of schemas.
    */
   private async processAll(): Promise<void> {
+    // processAll always has changes to process, so mark it as such for version tracking
+    this._sourceChangeDataState = "has-changes";
```
DanRod1999, line 2382: This is an actual change in behavior unlike the rest of the PR. I think this caught an actual "bug" in our code that we did not see because of the use of `_allowNoScopingESA`, but want to make sure this is not somehow breaking expected behavior.

Previously, when processAll() would occur, we would not update the syncVersion because the Data State would be "uninited". This is the default value set. Although we might want to reconsider this default setting in general, I think this fix makes the most sense for now.

However, since before this was "uninited" we would not update the sync version. I added a check in one of the imodel hub tests to show what the value should not be (it passes now because of this change). Since `processAll()` would incorrectly set the syncVersion, on the subsequent `processChanges()` the transformer would unnecessarily try to export entities when it doesn't need to (note: there is logic that prevents importing these elements I think, but we shouldn't be exporting either). I had a test that proved this but removed it as I think the check I added in an existing test proves the incorrect behavior enough. I can re add it or come up with a different test if we think it is necessary. Now that we set the sync version during `processAll()` workflows we don't redundantly go over exported elements that we don't care about on the next `process()` call.

JulijaRamoskiene: Tag this in PR as fixed: https://github.com/iTwin/imodel-transformer/issues/236
I know it was like this from very long ago, before I joined, We had to add workaround on our side. @mindaugasdirg FYI

common/api/imodel-transformer.api.md
```diff
@@ -259,47 +259,15 @@ export class IModelTransformer extends IModelExportHandler {
...
-        provenanceDb: IModelDb;
-        targetScopeElementId: Id64String;
-        isReverseSynchronization: boolean;
-        fn: (sourceElementId: Id64String, targetElementId: Id64String) => void;
-        skipPropagateChangesToRootElements: boolean;
-    }): Promise<void>;
-    // (undocumented)
-    getIsForwardSynchronization(): Promise<boolean>;
-    // (undocumented)
-    getIsReverseSynchronization(): Promise<boolean>;
```
JulijaRamoskiene, line 278: We use these methods: 

- `getProvenanceDb`
- 'getIsReverseSynchronization()'

Could you expose them in transformer?

DanRod1999: I guess you guys don't use this, but I think it makes sense to also expose in transformer too?
```typescript 
 public async getIsForwardSynchronization(): Promise<boolean> {
    return (await this.getSyncType()) === "forward";
  }
```

DanRod1999: exposed all three

JulijaRamoskiene (approved):

PR #296, opened 2026-05-12
WIP Refactor ECReferenceTypesCache

DanRod1999:
Modifications made by copilot from the substation team when trying to improve perf issues caught during placement of models. I'm creating this PR mostly so we can easily compare the changes.

Please see [issue](https://github.com/iTwin/imodel-transformer/issues/287) for some notes on perf

Issue #287, opened 2026-03-26
Improving ECReferenceTypeCache.ts


DanRod1999:
Users have been running into performance issues in ECReferenceTypeCache.ts, specifically during `initSchema()`.

There's been some proposed changes to the file to try to improve its performance.
- currently the [itwinjs-v5 ](https://github.com/iTwin/imodel-transformer/tree/itwinjs-v5) branch has some improvements
- more modifications have been made [here](https://github.com/iTwin/imodel-transformer/commit/24619efa1da03b553f73b1bd46257d42fd32adf6) using copilot to help.

Both changes add some fairly big modifications to the original file, and I tested some of them in app and in tests. They do give better perf, but I did notice that the hot path was really just `initSchema()`, so instead I used copilot to just modify `initSchema()` to see what perf I would get in my local tests, and it seemed to get most of the same perf gains (see file [here](https://github.com/iTwin/imodel-transformer/blob/dan/itwinjs-v5-optimized-init-schema/packages/transformer/src/ECReferenceTypesCache.ts)). While doing that I happened to test the difference in perf between debug and non debug runs:

### Non Debugger:
Trace | imodel-transformer.ECReferenceTypesCache | Completed 24 schemas in 317.36ms
### Debugger enabled:
Trace | imodel-transformer.ECReferenceTypesCache | Completed 24 schemas in 2078.26ms

### non debug:
Trace | imodel-transformer.ECReferenceTypesCache | Completed 24 schemas in 287.68ms 
### debug:
Trace | imodel-transformer.ECReferenceTypesCache | Completed 24 schemas in 745.06ms

### non debug
Trace | imodel-transformer.ECReferenceTypesCache | Completed 24 schemas in 364.50ms
### debug
Trace | imodel-transformer.ECReferenceTypesCache | Completed 24 schemas in 689.57ms 

Essentially I ended up seeing very little difference in performance between the original function vs the optimized versions wen not in debugger. Overall I think making perf improvements of any kind is good, but I am skeptical of changing a lot of this file if the gains are limited (modifying a lot of the file vs just editing `initSchema()`).

DanRod1999:
I created draft PR's with the different branches to better show the diffs:
[Only optimize `initSchema()`](https://github.com/iTwin/imodel-transformer/pull/297)
[Modify entire file](https://github.com/iTwin/imodel-transformer/pull/296)

DanRod1999:
[...]

PR #297, opened 2026-05-12
WIP Just optimize initSchema remove all caching optimizations

DanRod1999:
Only optimize 'initSchema()' function as that is where I caught most of the bottle neck in performance.

Opening this PR mostly just for the diff so it can be compared to this [PR](https://github.com/iTwin/imodel-transformer/pull/296).

Note, this file essentially resets ECReferenceTypesCache.ts to what is currently on main, but then adds some modifications to initSchema only. I think in general this minimizes the changes made to this file while keeping most of the perf gains. It is a bit slower than the "fully optimized file" however thats really only in debug mode (as mentioned in the [issue](https://github.com/iTwin/imodel-transformer/issues/287), and those changes I feel complicate the file more than necessary.

### DIFF
--- a/packages/transformer/src/ECReferenceTypesCache.ts
+++ b/packages/transformer/src/ECReferenceTypesCache.ts
@@ -11,6 +11,7 @@ import { ConcreteEntityTypes, RelTypeInfo } from "@itwin/core-common";
 import {
   ECClass,
   Mixin,
+  Property,
   RelationshipClass,
   RelationshipConstraint,
   Schema,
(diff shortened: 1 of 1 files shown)


hl662:
Can you update the pr title and description?

packages/transformer/src/ECReferenceTypesCache.ts
```diff
@@ -369,9 +302,5 @@ export class ECReferenceTypesCache {
   public clear() {
     this._initedSchemas.clear();
     this._propQualifierToRefType.clear();
```
hl662, line 304: `clear()` still needs to clear `_relClassNameEndToRefTypes` right? That cache is used to store the final computation from a relationship classname, to the RefTtypes? So if it isn't cleared, then you'll risk having a stale and incorrect cache?

hl662: maybe a test to check just to make sure?

DanRod1999: you were right, this should be cleared

packages/transformer/src/ECReferenceTypesCache.ts
```diff
@@ -143,38 +119,19 @@ export class ECReferenceTypesCache {
       SELECT s.Name as name
       FROM refs
       JOIN ECDbMeta.ECSchemaDef s ON refs.SchemaId=s.ECInstanceId
-      -- ensure schema dependency order
-      ORDER BY s.ECInstanceId
```
MichaelSwigerAtBentley, line 147: Why is this safeguard not needed anymore?

DanRod1999: Order should not be needed I believe. I based it on this https://github.com/iTwin/imodel-transformer/issues/273#issuecomment-4132603648, but maybe i misunderstood the context?

Issue #28, opened 2023-04-21
Error updating element when it gets remapped 


simsta6:
**Describe the bug**
![error updating elements, missing id drawio](https://user-images.githubusercontent.com/51540204/233617740-e4cf4063-a977-4389-8d62-016a65fa3773.png)

Source iModel changes, after first transformation, includes deleted displayStyle3D element and added displayStyle3D element with the same properties as the first one. 

onExportElement function remaps target DisplayStyle3D element. So now it points to newly added DisplayStyle3D (Id: 2 in source iModel).

When transformer reaches point where it deletes elements it deletes DisplayStyle3D element in target.

Then transformer calls finalizeTransformation and tries to update DisplayStyle3D element in target, because it has pending references (excludedElements in jsonProperties). It fails to update that element because it is deleted in target element

**To Reproduce**

Create empty iModel. Add DisplayStyle3d element that has excludedElements in jsonProperties.
Run transformation
Delete DisplayStyle3d element and add new DisplayStyle3d element with the same properties as first one (I think same code should be enough) in source iModel
Run transformation to trigger processChanges
Transformer crashes with 'error updating element: missing id'

**Expected behavior**
- It shouldn't crash
- Maybe transformer should create new element (with different id and federationGuid) instead of remapping that element

**Desktop (please complete the applicable information):**

OS: Windows
@itwin/imodel-transformer 0.1.2
iTwin.js Version 3.6.1

DanRod1999:
Resolved in linked branch, but we may look into better solutions in the future. 
Ideally we would like to consider having these types of changes be treated more like an update to a singular element instead of a delete and insert of 2 different elements. However, this is a more complicated change, so we will consider it more later if we find it necessary. I'll leave this issue open for now, but change its status.

PR #301, opened 2026-06-01
Remove from deleteids

DanRod1999:
Instead of storing a list of imported elements and then checking at time of delete if it is safe to delete the element, remove any elements from the deleteIds list that match with recently imported elements

Alternative solution to: https://github.com/iTwin/imodel-transformer/issues/28


DanRod1999:
Closing as a solution already exists and this was just an idea, but we believe current fix is better for now

PR #303, opened 2026-06-12
Fix for change operatinos when data is inserted into overflow table

JulijaRamoskiene:
Fix to change set parser to correctly handle scenario when elements are expanded into overflow table.

Related: https://github.com/iTwin/itwinjs-backlog/issues/2137


aruniverse:
@JulijaRamoskiene this should not be targeting main, but the v5 branch

DanRod1999:
I think the intent is we won't merge this until itwin-v5 branch is merged. Then this PR should be ok

aruniverse:
the commit history here is confusing then, and if you are waiting on that pr to merge then this needs to be drafted
you can target the v5 branch, and when that merges into main it will reparent this one

